PR #89, opened 2014-02-12
Stderr of job is redirected to stdout

MartinMystikJonas:
In test output show both output of stdout and stderr not only stdout because some errors are written to stderr.


dg:
Have you tried it on Windows? I think it doesn't work. `2>&1` works from command line, but not from `proc_open`.

JanTvrdik:
Would setting `bypass_shell` to `FALSE` help?

dg:
Yes, it helps. But bypass_shell must be enabled…

PR #133, opened 2014-05-24
Optional custom fail messages in asserts

MartinMystikJonas:
Sometimes test results can be hard to understand without detail examination of test code.

For example:
```PHP
Assert::true($obj->isA());
Assert::true($obj->isB());
```

You just get something like: 
```
   Failed: FALSE should be TRUE
   
   in Tester/Framework/Assert.php(370) 
   in Tester/Framework/Assert.php(141) Tester\Assert::fail()
   in tests/common/SomeTest.phpt(18) Tester\Assert::true()
```

Only information about what caused tha failure is line number on last line.

Also there are some cases when assertion is executed multiple times (for example asserting something for every item in collection) where we need to know for which item test failed.

Therefore it is usefull to have option to add custom fail messafe to asserts.

```PHP
Assert::true($obj->isA(), "Object is A");
Assert::true($obj->isB(), "Object is B");
```

And then you get:
You jsut get something like: 
```
   Failed: Object is B: FALSE should be TRUE
   
   in Tester/Framework/Assert.php(370) 
   in Tester/Framework/Assert.php(141) Tester\Assert::fail()
   in tests/common/SomeTest.phpt(18) Tester\Assert::true()
```

It could save you lot of time in examining test results.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -47,11 +47,11 @@ class Assert
 	 * Checks assertion. Values must be exactly the same.
 	 * @return void
 	 */
-	public static function same($expected, $actual)
+	public static function same($expected, $actual, $description = NULL)
 	{
 		self::$counter++;
 		if ($actual !== $expected) {
-			self::fail('%1 should be %2', $actual, $expected);
+			self::fail(self::describe('%1 should be %2', $description), $actual, $expected);
 		}
 	}
 
@@ -60,11 +60,11 @@ class Assert
 	 * Checks assertion. Values must not be exactly the same.
 	 * @return void
 	 */
-	public static function notSame($expected, $actual)
+	public static function notSame($expected, $actual, $description = NULL)
 	{
 		self::$counter++;
 		if ($actual === $expected) {
-			self::fail('%1 should not be %2', $actual, $expected);
+			self::fail(self::describe('%1 should not be %2', $description), $actual, $expected);
 		}
 	}
 
@@ -73,11 +73,11 @@ class Assert
 	 * Checks assertion. The identity of objects and the order of keys in the arrays are ignored.
 	 * @return void
 	 */
-	public static function equal($expected, $actual)
+	public static function equal($expected, $actual, $description = NULL)
 	{
 		self::$counter++;
 		if (!self::isEqual($expected, $actual)) {
-			self::fail('%1 should be equal to %2', $actual, $expected);
+			self::fail(self::describe('%1 should be equal to %2', $description), $actual, $expected);
 		}
 	}
 
@@ -86,11 +86,11 @@ class Assert
 	 * Checks assertion. The identity of objects and the order of keys in the arrays are ignored.
 	 * @return void
 	 */
-	public static function notEqual($expected, $actual)
+	public static function notEqual($expected, $actual, $description = NULL)
 	{
 		self::$counter++;
 		if (self::isEqual($expected, $actual)) {
-			self::fail('%1 should not be equal to %2', $actual, $expected);
+			self::fail(self::describe('%1 should not be equal to %2', $description), $actual, $expected);
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


MartinMystikJonas:
It this PR still considered as merge candidate?

fprochazka:
This feature is very usefull, I would like to see this merged. @MartinMystikJonas can you please rebase your pullrq?

MartinMystikJonas:
@fprochazka If there is real chance for this to be merged I will surely rebase it and write tests.

dg:
I think it is useful.

Tester/Framework/Assert.php
```diff
@@ -47,11 +47,11 @@ class Assert
 	 * Checks assertion. Values must be exactly the same.
 	 * @return void
 	 */
-	public static function same($expected, $actual)
+	public static function same($expected, $actual, $message = "")
```
dg: I know that `message` is commonly used, but would not it be more accurate to use `description`? 

Btw, use please single quotes where possible.

MartinMystikJonas: I can rename it but message is common name in this context.

I will change it to single quotes

Tester/Framework/Assert.php
```diff
@@ -60,11 +60,11 @@ public static function same($expected, $actual)
...
 	 * @return void
 	 */
-	public static function notSame($expected, $actual)
+	public static function notSame($expected, $actual, $message = "")
 	{
 		self::$counter++;
 		if ($actual === $expected) {
-			self::fail('%1 should not be %2', $actual, $expected);
+			self::failMessage($message, '%1 should not be %2', $actual, $expected);
 		}
```
dg: What about simple `self::fail(($message ? "$message: " : '') . '%1 should not be %2', $actual, $expected);` or
`self::fail(self::join('%1 should not be %2', $message), $actual, $expected);`?

MartinMystikJonas: I preferr this solution. It will be much simpler to change format of output with message in future.

MartinMystikJonas:
Rebased, fixed and tests added.

Tester/Framework/Assert.php
dg: I do not understand, can you explain it?

MartinMystikJonas: If we deside for example to add colors to custom message change of specialized method failMessage will be simpler I think. Because it could be something more complex than simple join in future

milo:
:+1: One more thing, update commit message please to be consistent.

I'm thinking, if a longer backtrace has some influence for somebody, but probably not.

dg:
failMessage is really bad name. Is new method really needed?

milo:
Idea: what about passing `array($description, '....')` to fail and concat here?

MartinMystikJonas:
From my POW separate this to method is better. It is only private so it will not change signature of class and I think it improves readability. Maybe rename it to `failWithDescription`?

vojtech-dobes:
Maybe stupid idea, but if some method is necessary, what about `failBecause`?

MartinMystikJonas:
@milo: What change in commit message do you mean?

Tester/Framework/Assert.php
dg: So in future can join() do something more complex than simple join.

MartinMystikJonas: I do not like when method do something more than it name suggests. But if you really want this change I will do it

dg: Of course, name `join` is just suggestion. 

I am just trying to somehow solve problematic `failMessage`. We can rename it to failWithDescription (too ugly), we can add 4th argument to fail() (too weird), etc. 

Btw: to add color will require much more changes in code.

dg:
Conversation about failMessage is here https://github.com/nette/tester/pull/133#discussion_r36080522 but is hidden in outdated diff.

Tester/Framework/Assert.php
MartinMystikJonas: Ok what about this:

```
self::fail(self::message('%1 should not be %2', $description), $actual, $expected);

private message ($reason, $description) {
  // ...
}
```

milo:
@dg I see!

@MartinMystikJonas 
Commit message like `Assert: added custom fail description`
And as you wrote, private `self::message($reson, $description)` seems OK to me.

dg:
Verb is better. What about describe() ?

MartinMystikJonas:
@dg: changed to descibe()

@milo: Commit message updated

dg:
:+1:

milo:
@MartinMystikJonas please, add tests for `notSame, notEqual, contains, falsey` and :100:

MartinMystikJonas:
done

dg:
Thanks

milo:
@MartinMystikJonas Thank you!

PR #134, opened 2014-05-25
Dumper::dumpException() writes line of source code

dg:
Partially solves this  #133

### DIFF
--- a/src/Framework/Dumper.php
+++ b/src/Framework/Dumper.php
@@ -321,8 +321,10 @@ class Dumper
 					)
 					: '[internal function]'
 				)
-				. $item['class'] . $item['type']
-				. (isset($item['function']) ? $item['function'] . '()' : '')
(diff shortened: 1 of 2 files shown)


dg:
@milo what about this approach?

Before:
image

After:
image

src/Framework/Dumper.php
```diff
@@ -321,8 +321,10 @@ public static function dumpException($e)
 					)
 					: '[internal function]'
 				)
-				. $item['class'] . $item['type']
-				. (isset($item['function']) ? $item['function'] . '()' : '')
+				. ($item['class'] === 'Tester\Assert' && ($tmp = file($item['file'])) && strpos($tmp[$item['line'] - 1], ":$item[function](")
+					? trim($tmp[$item['line'] - 1])
```
dg: In PHP 5.4 can be removed $tmp

milo:
@dg :+1:

PR #187, opened 2015-01-03
Added fileExists assertion (Closes #186)

### LINKED ISSUE #186: Add Assert::fileExists() assertion
I'm trying Tester for the first time (I'm using PHPUnit) in a project and I miss this assertion.

I understand that Tester is trying to have only the most important assertions. Also the need of "missing" assertion could differ from the project type or user.

I just don't like the code and its "fail" message:
```php
$isFile = file_exists('somefile');
Assert::true($isFile);
```
I could misuse the Assert::matchFile(), but that doesn't seem right.

For those asking the possible question "why the hell do you need to check file existence without checking file content": TDD and testing a class that creates directory structure based on given parameters have leaded me to this issue.

Are there any strong arguments against?

BigCharlie:
I don't know the strategy about adding new assertions for Tester. I've needed this one, but others may not use it.

### DIFF
--- a/Tester/Framework/Assert.php
+++ b/Tester/Framework/Assert.php
@@ -417,6 +417,18 @@ class Assert
 
 
 	/**
+	 * Checks file existence
+	 * @return void
+	 */
+	public static function fileExists($file) {
+		self::$counter++;
+		if (!file_exists($file)) {
+			self::fail('File %1 does not exist', $file);
+		}
+	}
+
+
+	/**
 	 * Failed assertion
 	 * @return void
 	 */
--- a/Tester/Framework/Assert.php
+++ b/Tester/Framework/Assert.php
@@ -420,9 +420,10 @@ class Assert
 	 * Checks file existence
 	 * @return void
 	 */
-	public static function fileExists($file) {
+	public static function fileExists($file)
+	{
 		self::$counter++;
-		if (!file_exists($file)) {
+		if (!is_file($file)) {
 			self::fail('File %1 does not exist', $file);
 		}
 	}
--- a/readme.md
+++ b/readme.md
@@ -96,6 +96,7 @@ This table shows all assertions (class `Assert` means `Tester\Assert`):
 - `Assert::match($pattern, $value)` - Compares result using regular expression or mask.
 - `Assert::matchFile($file, $value)` - Compares result using regular expression or mask sorted in file.
 - `Assert::count($count, $value)` - Reports an error if number of items in $value is not $count.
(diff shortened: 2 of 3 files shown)


JanTvrdik:
1. Commit message should start with `Assert: `
2. "Closes …" should be in square brackets
3. The coding style for method brackets is wrong.
4. The unit test should not use `rand()`
5. I would prefer using `is_file()` instead of `file_exists`

Vrtak-CZ:
> I would prefer using `is_file()` instead of `file_exists`

Why?

JanTvrdik:
Because it is rare that you would not know whether the target path is directory or file.

Vrtak-CZ:
> We may consider adding `Assert::dirExists()` as well.

:+1:

matej21:
:-1: for adding these assertions (fileExists and dirExists) to the Assert class. We may consider adding new classes like "IOAssert" (or e.g. DatabaseAssert)... but I think, Assert class should handle only generic assertions. 

Anyway, these assertions could be solved in the userland. Both Assert::$counter and Assert::fail are public. So it could be solved as a separate package.

fprochazka:
In my opinion, this is perfectly fine with me

```
Assert::true(is_file($file))
```

but I'm not agains adding this.

milo:
@BigCharlie Thank you for the PR! I understand the "fail message" problem, but in my opinion, adding such method(s) is not necessary. You can:
```php
if (!is_file($file)) {
    Assert::fail(...); 
}
// or just
is_file($file) || Assert::fail(...); 
```

or you can extend the Assert class if you use it very often in your tests (beware of future BC):
```php
class Assert extends Tester\Assert
{
    public static function file($file) {...}
}
```

Maybe extend the [Assert::type()](https://github.com/nette/tester/blob/76153d57bd1175142e4fa800e8782a7449d11136/Tester/Framework/Assert.php#L246) method for these...

BigCharlie:
Thanks all for feedback. I would not expect so many mistakes in couple of lines of code. I would try better next time.

@milo thanks for other examples, that could work for me. Your last proposal of extending Assert:type() method seems confusing for me:
- method description would contain to many "or"
- I would not expect file assertion in this method

BigCharlie:
> The unit test should not use rand()

@JanTvrdik How should I solve it - would *uniqid* be ok? Or all the test code should be deterministic, so some 'thisFileNotExists' should make it?

JanTvrdik:
> Or all the test code should be deterministic, so some 'thisFileNotExists' should make it?

Yes, if possible all tests should avoid any randomness.

PR #190, opened 2015-01-04
Runner: interrupted may be set in test initiation phase

milo:
The `Runner::$interrupted` is reseted to FALSE because it is needed in a `--watch` mode. But once per run() is enough.

I hope I didn't forget for some case, it is difficult to create a test case for this behaviour.


milo:
Added tests for `--stop-on-fail`.

hrach:
:+1:

PR #192, opened 2015-01-08
tests: added non-blocking Job test [WIP]

dg:
I wanted to add the test for `run(FALSE)`, but it works only in CLI mode. In CGI mode, I must call `stream_get_contents` after process is closed, it seems, that reading is blocking… Why?

The `$this->output .= stream_get_contents($this->stdout);` was added by @vrana here e641f6a9d81fef5dc9b68a8d63cf7648331005e4, but I don't know if it's really needed.

### DIFF
--- a/src/Runner/Job.php
+++ b/src/Runner/Job.php
@@ -168,6 +168,7 @@ class Job
 			return true;
 		}
 
+		$this->output = stream_get_contents($this->stdout);
 		fclose($this->stdout);
 		if ($this->stderr) {
 			fclose($this->stderr);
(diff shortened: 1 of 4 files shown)


JanTvrdik:
Maybe because it does not work on Windows? http://php.net/manual/en/function.stream-set-blocking.php#110997

milo:
It is blocking. The `stream_set_blocking()` returns FALSE, the `stream_get_meta_data()` informs that it is blocked and found this [PHP BUG](https://bugs.php.net/bug.php?id=47918). But I don't know why it works with CLI, the dumped info is the same.

dg:
Teď teda vůbec nechápu, jak může Tester fungovat? :-O

PR #193, opened 2015-01-08
Job: added getErrorOutput() [WIP]

### LINKED ISSUE #86: Show stderr in test output
Is there reason why tester shows only content of stdout of test execution after test failure?

I just spent about hour trying to figure out why my tests suddenly failing with no information at all, because causing exception was written to stderr.

It should be sufficient to append "2>&1" to Job command at Job.php:77

Or is there reason for not doing so?

dg:
It seems that it works for big error output when errorOutput is readed this way. See https://github.com/nette/tester/issues/86#issuecomment-35213229

 [Closes #86][Closes #89]

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -49,8 +49,10 @@ class CliTester
 
 		if ($this->options['--info']) {
 			$job = new Job(__DIR__ . '/info.php', $this->interpreter);
-			$job->run();
+			$job->run($job::RUN_COLLECT_ERRORS);
 			echo $job->getOutput();
+			echo $job->getErrorOutput() ? \Tester\Dumper::color('red', 'Errors:') . "\n" . $job->getErrorOutput() : '';
+			echo "\n\n";
 			return;
 		}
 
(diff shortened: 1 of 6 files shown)


milo:
This is almost the identical commit as I wrote in February https://github.com/milo/tester/commit/02daf6a047d7246d10a255a2594b97f99ce72853 :-)

I tested it now and it works. I don't know what exactly was wrong, probably in PHP 5.3. So :+1:

dg:
There is a huge difference :-) 

```
$this->output .= stream_get_contents($this->stdout);
$this->errorOutput .= stream_get_contents($this->stderr);
```

versus 

```
$this->errorOutput .= stream_get_contents($this->stderr);
$this->output .= stream_get_contents($this->stdout);
```

JanTvrdik:
Could you explain why does THAT make it work?

dg:
I have no idea

milo:
Well, that's the last state only :-) I swear I wrote it in the other order, up side down and in japanese and still was buggy! :)

dg:
I am not sure that it works. I'd like to test it few days before merge.

PR #194, opened 2015-01-08
info.php: shorter output

dg:
I feel that `--info` output could be more concentrated...


milo:
Probably yes, but seems to me to much concentrated when without colors.

dg:
Thats right, what about this?

milo:
:+1: for merge

PR #201, opened 2015-01-22
Allow to acquire multiple locks

MartinMystikJonas:
In current implementation when Environment::lock is called multiple times only last acquired lock is saved. Previous lock are sometimes randomly garbage collected. This results in race conditions and random test failures.

### DIFF
--- a/Tester/Framework/Environment.php
+++ b/Tester/Framework/Environment.php
@@ -136,8 +136,8 @@ class Environment
 	 */
 	public static function lock($name = '', $path = '')
 	{
-		static $lock;
-		flock($lock = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
+		static $locks;
+		flock($locks[] = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
 	}
 
 


JanTvrdik:
Why do you need multiple locks?

MartinMystikJonas:
In my case one for database (to prevent parallel accesses to database) and second for testing API server (also to prevent parellel access to API). Some tests need just database, some needs just API but some needs both.

JanTvrdik:
Ok, that makes sense.

Tester/Framework/Environment.php
```diff
@@ -136,8 +136,8 @@ public static function skip($message = '')
 	 */
 	public static function lock($name = '', $path = '')
 	{
-		static $lock;
-		flock($lock = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
+		static $locks = array();
```
milo: @MartinMystikJonas Can you remove the initialization? No big deal, it's just shorter.

milo:
@MartinMystikJonas And can you rename the commit? Something like `Environment: allow multiple locks...` I'll merge it.

MartinMystikJonas:
@milo Done.

milo:
@MartinMystikJonas Thank you!

mrtnzlml:
Hi. Are you really sure this is a good idea? I don't know the original use case, but with this commit is even very simple test failing (based on sandbox default test case):

```php
<?php

namespace Test;

use Nette,
	Tester,
	Tester\Assert;

$container = require __DIR__ . '/bootstrap.php';

class ExampleTest extends Tester\TestCase {

	private $container;

	function __construct(Nette\DI\Container $container) {
		$this->container = $container;
	}

	function setUp() {
		Tester\Environment::lock('db', __DIR__);
	}

	function testSomething1() {
		Assert::true( true );
	}

	function testSomething2() {
		Assert::true( true );
	}

}

$test = new ExampleTest($container);
$test->run();
```

Well it's actually not failing, it freezes...

fprochazka:
This is wrong. You should use different names.

Tester/Framework/Environment.php
```diff
@@ -136,8 +136,8 @@ public static function skip($message = '')
 	 */
 	public static function lock($name = '', $path = '')
 	{
-		static $lock;
-		flock($lock = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
+		static $locks;
+		flock($locks[] = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
```
fprochazka: This should have been `$locks[$name]`

fprochazka:
The lock is never released, because you're not running the test case in multiple processes. To run the test case in multiple processes you have to add `@testCase` annotation above the `namespace`. That said, I think this implementation is wrong.

mrtnzlml:
It's not true. I have every testcase with this annotation. This example also doesn't work with the `@testCase`. So you are saying my implementation is wrong? I have to use different lock names for each test? Doesn't make much sense to me.

milo:
@mrtnzlml Multiple locks is desired feature. E.g. `lock('db'); lock('api');`. You should call `lock()` ouside the TestCase to prevent freezing. But I understand it is unexpected.

See #210

milo:
@mrtnzlml Are you sure you use `@testCase` in very first file annotation? Imho, it should not freeze with `@testCase` because every `test....` method runs in own process.

mrtnzlml:
@milo Yeah, I understood the request and it make sense. But the implementation was odd and I wasn't able to run all tests without freeze.

Oh sorry. It actually works with this annotation. It doesn't work only in my project (that's why I reported this behavior). But to be honest I am not sure if it actually run in 8 threads or not. Apparently not.

@fprochazka What did you mean with _"You should use different names."_?

milo:
Easy to check. Run `vendor/bin/tester path/to/one/Test.phpt` and you should get as many results as you have test methods.

mrtnzlml:
Yeah I know that and it returns more dots (which is not very good test :-)). But still it doesn't work with the annotation. But I am not very good in testing and it works well with your patch, so - problem solved I guess. Thanks.

fprochazka:
@mrtnzlml why are you trying to solve what I've meant by comment that I've deleted?

PR #210, opened 2015-02-21
Environment: acquire lock() only if not exist

milo:
Contents of static $locks exists only in one process. The acquiring can be skipped when lock already exists in this process.

### DIFF
--- a/Tester/Framework/Environment.php
+++ b/Tester/Framework/Environment.php
@@ -132,13 +132,18 @@ class Environment
 
 
 	/**
-	 * locks the parallel tests.
+	 * Locks the parallel tests.
+	 * @param  string
+	 * @param  string  lock store directory
 	 * @return void
 	 */
 	public static function lock($name = '', $path = '')
 	{
 		static $locks;
-		flock($locks[] = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
+		$file = "$path/lock-" . md5($name);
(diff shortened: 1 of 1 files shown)


fprochazka:
:+1:

mrtnzlml:
Looks good. Tested.

Tester/Framework/Environment.php
```diff
@@ -132,13 +132,18 @@ public static function skip($message = '')
...
 	 * @return void
 	 */
 	public static function lock($name = '', $path = '')
 	{
 		static $locks;
-		flock($locks[] = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
+		$file = "$path/lock-" . md5($name);
+		if (!isset($locks[$file])) {
+			flock($locks[$file] = fopen($file, 'w'), LOCK_EX);
+		}
```
JanTvrdik: Will not simply

~~~php
flock($locks[$name] = fopen($path . '/lock-' . md5($name), 'w'), LOCK_EX);
~~~

milo:
It releases previous lock for a few microseconds.

Tester/Framework/Environment.php
milo: It releases previous lock for a few microseconds.

milo:
@dg Some comments on this?

dg:
It seems ok.

PR #196, opened 2015-01-12
CodeCoverage Clover XML

milo:
This is still WIP. I would appreciate if somebody can test it with tools like Jenkins CI.

To try it, call Tester as usually, but set the coverage file extension to xml like:
```
tester --coverage coverage.xml
```

PR #211, opened 2015-02-22
CodeCoverage CloverXML (robust) [WIP]

milo:
The #196 introduced very stripped Clover XML code coverage output. This solution is more complete.

The CloverXMLGenerator tries to mimic the PhpUnit output (not completely, yet), because there is no official CloverXML for PHP specification (e.g. [this reply](https://youtrack.jetbrains.com/issue/WI-26350) from phpStorm).

To reach that, source code must be analysed. Class `PhpParser` is introduced, it scans PHP code for classes/traites/interfaces and their methods start/end lines. Big amount of code is stolen from [Nette\Reflection\AnnotationsParser](https://github.com/nette/reflection/blob/08328bbd23323de285966dac7164f9ceb73cd77b/src/Reflection/AnnotationsParser.php#L271).

The desing is very young and there are questions:
- `PhpParser` can be moved into `Helpers::parsePhpCode()`
- <del>should be `PhpParser::output()`, currenty `stdClass`, in own object?</del>
- <del>is the DOM necessarry?</del>
- support standalone functions?
- any better naming is welcomed
- <del>maybe somehow replace PhpParser by Reflection in Collector</del> (cannot, must count commented lines)
- do it more friendly for future report generators?

Questions about implementation are often as `// TODO:`, but:
- what about interfaces?
- what about abstract methods?
- prepare for XDebug [finding paths](http://derickrethans.nl/path-branch-coverage.html)? :)

Even some missing elements attributes, tests with 3rd party tools are very welcomed.


milo:
Ref @Vrtak-CZ @o5

Vrtak-CZ:
After few commits (https://github.com/Vrtak-CZ/tester/compare/clover) I have working version tested with Jenkins and Coveralls (https://github.com/satooshi/php-coveralls).

milo:
@Vrtak-CZ Thank you! That's great to know which elements and attributes are mandatory.

milo:
I'm thinking about another approach. Enclose whole analysis into `CloveredFileInfo`, e.g.:
```php
class CoveredFileInfo
{
    public function __construct($fileName, array $fileCoverageData) {...}
    public function getClasses() {...}
    public function getMethods($class) {...}
    public function getLineCount() {...}
    public function getNonCommentedLineCount() {...}
    ...
}
```
It is better interface for another posible output format (it can be used by current HTML output too). But current state can be easily switched to this, so maybe in next version if necessary.

TomasVotruba:
@Vrtak-CZ Could you try also **Scrutinizer** (https://scrutinizer-ci.com)?

Vrtak-CZ:
@TomasVotruba looks like working https://scrutinizer-ci.com/g/nella/forms-datetime/inspections/4e98187f-8bff-47ab-a19e-f169df7153e5

TomasVotruba:
@Vrtak-CZ Thanks. Great job!

o5:
Does anybody try to run nette/tester coverage under Travis? 

From Travis [documentation](http://docs.travis-ci.com/user/languages/php/#Preinstalled-PHP-extensions), Xdebug is enabled by default but nette/tester [throws error](https://travis-ci.org/o5/grido/builds/52532871#L288):
<code>Error: Code coverage functionality requires Xdebug extension (used php)</code>

I've tried to debug and I found probably an unnecessary param <code>-n</code> (<code>No php.ini file will be used</code>) [added](https://github.com/nette/tester/commit/32122b7f669fafeb3190536f84cc761e8ff4b5f0) by milo. So I used my fork of nette/tester, but [here's](https://github.com/nette/tester/blob/master/Tester/CodeCoverage/Collector.php#L28) the second check. 

I don't have any idea, why it is doesn't work, because [here](https://travis-ci.org/o5/grido/builds/52527910#L277-L278) as you can see Xdebug is successfully loaded. 

Any tips?
Thanks.

milo:
@o5 Never tried code coverage on Travis. The `-n` is necessary to clean a testing environmet. Try:

```
vendor/bin/tester -d extension=xdebug.so --info
```
to check Xdebug will be loaded for tests.

Vrtak-CZ:
So is there anything what I can do?

milo:
@Vrtak-CZ I merged your commits content and performed some refactoring. Can you test it again?

milo:
@dg May I ask you for review?

milo:
Two things I know about left:
- `CliTester::prepareCodeCoverage()` should return generator (possible Exception, e.g. missing DOM extension, prevents tests run)
- test how PhpParser behaves with abstract methods :)

Vrtak-CZ:
Tested with Jenkins & Coveralls. Works! thx @milo

milo:
@dg This is, imho, ready to be merged.

dg:
Wow, that's a pretty complex.

dg:
It is intentional that AbstractGenerator::render() can produce partially generated files?

Tester/CodeCoverage/PhpParser.php
```diff
@@ -0,0 +1,183 @@
...
+	 * where $methodInfo is:
+	 *     stdClass {
+	 *         start: int,
+	 *         end: int,
+	 *         visibility: public|protected|private
+	 *     }
+	 */
+	public function parse($code)
+	{
+		$tokens = @token_get_all($code);
```
dg: // @ - source code can be written in newer PHP

Tester/CodeCoverage/PhpParser.php
```diff
@@ -0,0 +1,183 @@
...
+			'linesOfCode' => max(1, substr_count($code, "\n")),
+			'linesOfComments' => 0,
+			'functions' => array(),
+			'classes' => array(),
+			'traits' => array(),
+			'interfaces' => array(),
+		);
+
+		$T_TRAIT = PHP_VERSION_ID < 50400 ? -1 : T_TRAIT;
+		while(list(, $token) = each($tokens)) {
```
dg: missing space after while

Tester/CodeCoverage/PhpParser.php
```diff
@@ -0,0 +1,183 @@
...
+				}
+				$line = $token[2];
+			}
+
+			switch (is_array($token) ? $token[0] : $token) {
+				case T_NAMESPACE:
+					$namespace = ltrim(self::fetch($tokens, array(T_STRING, T_NS_SEPARATOR)) . '\\', '\\');
+					break;
+
+				case T_CLASS:
```
dg: Is it ready for syntax `ClassName::class`?

Tester/CodeCoverage/Generators/AbstractGenerator.php
```diff
@@ -0,0 +1,116 @@
...
+		ob_end_flush();
+
+		if (isset($e)) {
+			throw $e;
+		}
+	}
+
+
+	/**
+	 * @return \SplFileInfo[]
```
dg: @return AcceptIterator?

milo:
> Wow, that's a pretty complex.

Agree. I take any proposes to simplyfied it. (e.g. PhpParser as part of Helpers?)

> It is intentional that AbstractGenerator::render() can produce partially generated files?

I don't understand what you mean. I took logic from [current ReportGenerator](https://github.com/nette/tester/blob/f3bf32d831dc462a66d9a115f4345397335e658b/Tester/CodeCoverage/ReportGenerator.php#L81).

Tester/CodeCoverage/PhpParser.php
milo: @dg I guess yes, the `self::fetch(..., T_STRING)` condition follows. But I'll add it to tests.

dg:
I mean the code that catches exceptions. It resets output buffer, but leaves file only partially generated.

milo:
Hm, looks like workaround to me now. The only exception can be `Missing DOM extension` from CloverXMLRepor constructor which should be thrown before tests run.

Thank you for the review, I'll fix it.

milo:
@dg All typos fixed. About an exception in the `AbstractGenerator::reder()`... I added this 11b5676f4a419e92049aee6c964ab7d5284a9bd3. Catching ensures, that exception trace is not stored into rendered output.

Tester/CodeCoverage/Generators/AbstractGenerator.php
```diff
@@ -0,0 +1,118 @@
...
+		$handle = $file ? @fopen($file, 'w') : STDOUT;
+		if (!$handle) {
+			throw new \Exception("Unable to write to file '$file'.");
+		}
+
+		ob_start(function($buffer) use ($handle) { fwrite($handle, $buffer); }, 4096);
+		try {
+			$this->renderSelf();
+		} catch (\Exception $e) {
+			ftruncate($handle, 0);
```
dg: Maybe ob_end_flush can write to file, so what about to move it to `if (isset($e))`? Or close file after ob_end_flush() and unlink it in  `if (isset($e))`?

Tester/CodeCoverage/Generators/CloverXMLGenerator.php
```diff
@@ -0,0 +1,213 @@
+<?php
+
+/**
+ * This file is part of the Nette Tester.
+ * Copyright (c) 2009 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Tester\CodeCoverage\Generators;
+
+use \DOMDocument,
```
dg: Leading `\` is unnecessary.

Tester/CodeCoverage/Generators/CloverXMLGenerator.php
```diff
@@ -0,0 +1,213 @@
...
+				$elClassMetrics = $elClass->appendChild($doc->createElement('metrics'));
+				$classMetrics = $this->calculateClassMetrics($info, $coverageData);
+				self::setMetricAttributes($elClassMetrics, $classMetrics);
+				self::appendMetrics($fileMetrics, $classMetrics);
+			}
+			self::setMetricAttributes($elFileMetrics, $fileMetrics);
+
+
+			foreach ((array) $coverageData as $line => $count) {
+				if ($count === -2) { // -2 = dead code
```
dg: What about to use constants, ie. AbstractGenerator::CODE_DEAD, CODE_TESTED and CODE_UNTESTED?

dg:
Great!

Tester/CodeCoverage/Generators/AbstractGenerator.php
milo: Good point. Unlink seems better since empty file can be invalid contents.

Tester/CodeCoverage/Generators/CloverXMLGenerator.php
milo: Pfmpf. Blaming phpStorm :)

Tester/CodeCoverage/Generators/CloverXMLGenerator.php
milo: :+1:

milo:
Updated. So, if there will be no more suggestions, I'll merge it tomorrow when Travis will be merciful.

Vrtak-CZ:
:heart: :clap:

dg:
:+1: 

„A přitom taková blbost“ :)

PR #212, opened 2015-03-11
Assert: equal does not use strict comparison

vysinsky:
I think `Assert::equal()` should not use strict comparison `===` but only weak comparison `==` should be used according to phpdoc `The identity of objects and the order of keys in the arrays are ignored.`

### DIFF
--- a/Tester/Framework/Assert.php
+++ b/Tester/Framework/Assert.php
@@ -524,7 +524,7 @@ class Assert
 			return TRUE;
 		}
 
-		return $expected === $actual;
+		return $expected == $actual;
 	}
 
 }
--- a/tests/Framework/Assert.equal.phpt
+++ b/tests/Framework/Assert.equal.phpt
@@ -27,7 +27,12 @@ $float1 = 1 / 3;
 $float2 = 1 - 2 / 3;
 
 $equals = array(
+	array(1, 1.0),
 	array(1, 1),
+	array(1, "1"),
+	array(1, TRUE),
+	array(0, FALSE),
+	array(0, NULL),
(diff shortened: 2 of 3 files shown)


dg:
:-1:

vysinsky:
Ok, so how to compare two values but not strictly? E.g. 1 and "1" (for example from a database)?

matej21:
`Assert::same`

matej21:
`Assert::ok(1 == $bar)` ?

vysinsky:
@matej21 yes it is possible way, but the same way I can do `Assert::true(1 === 1)`.

dg:
What about `Assert::equal(1, ' 1 ahoj')`, should it be equal?

vysinsky:
No. You got me :)

dg:
We can discuss about int + string. But weak comparsion is too week.

vysinsky:
Well, now I see this change would be BC break. And not as useful as I thought. For comparing int and string one can use `Assert::true(1 == "1")`.

dadyday:
I understand, that int and string comparisation may not result in true, but whats about int and float values? Assert::equal(1, 1.0) fails ... :/

its eventually possible to add a weak mathematical test:
Assert::nearly(2, 1.9999999999999999) => true (within a definable precision)

milo:
Type juggling is always problematic, but `Assert::equal()` behaviour is correct. It detects change of type (from `int` to `float` for example) which is important in `strict_mode=1`.

Comparison of floats by difference is implemented. Your example can be compared as:
```php
Assert::equal(2.0, 1.9999999999999999);
```

PR #215, opened 2015-03-14
Added support for generating test report in JUnit XML format

vysinsky:
- feature
- BC break - no
- documentation - not needed

JUnit format specification: http://help.catchsoftware.com/display/ET/JUnit+Format
For example [CircleCI](https://circleci.com/) supports test reports in this format. It shows reports nicely in the GUI.

image

### DIFF
--- a/Tester/Runner/CliTester.php
+++ b/Tester/Runner/CliTester.php
@@ -36,7 +36,7 @@ class CliTester
 		Environment::$debugMode = (bool) $this->options['--debug'];
 		if (isset($this->options['--colors'])) {
 			Environment::$useColors = (bool) $this->options['--colors'];
-		} elseif ($this->options['-o'] === 'tap') {
+		} elseif (in_array($this->options['-o'], array('tap', 'junit'))) {
 			Environment::$useColors = FALSE;
 		}
 
@@ -96,21 +96,21 @@ Usage:
     tester.php [options] [<test file> | <directory>]...
 
 Options:
-    -p <path>              Specify PHP interpreter to run (default: php-cgi).
-    -c <path>              Look for php.ini file (or look in directory) <path>.
-    -l | --log <path>      Write log to file <path>.
-    -d <key=value>...      Define INI entry 'key' with value 'val'.
-    -s                     Show information about skipped tests.
-    --stop-on-fail         Stop execution upon the first failure.
-    -j <num>               Run <num> jobs in parallel (default: 8).
-    -o <console|tap|none>  Specify output format.
-    -w | --watch <path>    Watch directory.
-    -i | --info            Show tests environment info and exit.
-    --setup <path>         Script for runner setup.
-    --colors [1|0]         Enable or disable colors.
-    --coverage <path>      Generate code coverage report to file.
-    --coverage-src <path>  Path to source code.
-    -h | --help            This help.
+    -p <path>                    Specify PHP interpreter to run (default: php-cgi).
+    -c <path>                    Look for php.ini file (or look in directory) <path>.
+    -l | --log <path>            Write log to file <path>.
+    -d <key=value>...            Define INI entry 'key' with value 'val'.
+    -s                           Show information about skipped tests.
+    --stop-on-fail               Stop execution upon the first failure.
+    -j <num>                     Run <num> jobs in parallel (default: 8).
+    -o <console|tap|junit|none>  Specify output format.
+    -w | --watch <path>          Watch directory.
+    -i | --info                  Show tests environment info and exit.
+    --setup <path>               Script for runner setup.
+    --colors [1|0]               Enable or disable colors.
+    --coverage <path>            Generate code coverage report to file.
+    --coverage-src <path>        Path to source code.
+    -h | --help                  This help.
 
 XX
 		, array(
@@ -184,9 +184,16 @@ XX
 		$runner->stopOnFail = $this->options['--stop-on-fail'];
 
 		if ($this->options['-o'] !== 'none') {
-			$runner->outputHandlers[] = $this->options['-o'] === 'tap'
-				? new Output\TapPrinter($runner)
-				: new Output\ConsolePrinter($runner, $this->options['-s']);
+			switch ($this->options['-o']) {
+				case 'tap':
+					$runner->outputHandlers[] = new Output\TapPrinter($runner);
+					break;
+				case 'junit':
+					$runner->outputHandlers[] = new Output\JUnitPrinter($runner);
+					break;
+				default:
+					$runner->outputHandlers[] = new Output\ConsolePrinter($runner, $this->options['-s']);
+			}
 		}
 
 		if ($this->options['--log']) {
@@ -222,7 +229,7 @@ XX
 	/** @return void */
 	private function finishCodeCoverage($file)
 	{
-		if ($this->options['-o'] !== 'none' && $this->options['-o'] !== 'tap') {
+		if (!in_array($this->options['-o'], array('none', 'tap', 'junit'), TRUE)) {
 			echo "Generating code coverage report\n";
 		}
 		$generator = new CodeCoverage\ReportGenerator($file, $this->options['--coverage-src']);
--- a/Tester/Runner/Output/JUnitPrinter.php
+++ b/Tester/Runner/Output/JUnitPrinter.php
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * This file is part of the Nette Tester.
+ * Copyright (c) 2009 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Tester\Runner\Output;
+
+use Tester,
+	Tester\Runner\Runner;
+
+
+/**
+ * JUnit xml format printer.
+ */
+class JUnitPrinter implements Tester\Runner\OutputHandler
+{
+	/** @var Runner */
+	private $runner;
+
+	/** @var resource */
+	private $file;
+
+	/** @var string */
+	private $buffer;
+
+	/** @var float */
+	private $startTime;
+
+	public function __construct(Runner $runner, $file = 'php://output')
+	{
+		$this->runner = $runner;
+		$this->file = fopen($file, 'w');
+	}
+
+
+	public function begin()
+	{
+		$this->startTime = microtime(TRUE);
+		fwrite($this->file, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<testsuites>\n");
+	}
+
+
+	public function result($testName, $result, $message)
+	{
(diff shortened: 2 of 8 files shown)


Tester/Runner/Output/JUnitWriter.php
```diff
@@ -0,0 +1,78 @@
...
+	public function __construct(Runner $runner, $file = 'php://output')
+	{
+		$this->runner = $runner;
+		$this->file = fopen($file, 'w');
+	}
+
+
+	public function begin()
+	{
+		$this->time = -microtime(TRUE);
```
Majkl578: This doesn't make much sense as a *time* variable, why do you need `-` here anyway?

vysinsky: I took it from https://github.com/nette/tester/blob/master/Tester/Runner/Output/ConsolePrinter.php#L42-42

fprochazka: instead of subtracting ot at the end, he just sums those values, because the beginning value is negative :)

Majkl578: I understand that, but then the variable name `time` doesn't make sense. :) And there is no benefit of doing this, or is there?

Majkl578: Actually, performance-wise this solution appears to be worse, it results in one extra opcode:
http://3v4l.org/elISt/vld#tabs
vs.
http://3v4l.org/sgeKV/vld#tabs

fprochazka: Someone probably thought this was a nice hack. But I don't see the benefit either :)

vysinsky: Ok. Fixed :)

milo:
@vysinsky I agree about adding `JUnit` output format, it is supported by many tools. About implementation...

Do you need to write the JUnit output into file? If don't and stdout is enough, there is an `-o` option which can be extended.

vysinsky:
Do you want to output generated xml into console and possibly save it with '> out.xml'? 
Yeah maybe it will be better solution than adding new parameter `--junit`

milo:
@vysinsky Yes, imho it's better.

vysinsky:
@milo I agree. I will change it soon.

Tester/Runner/Output/JUnitWriter.php
```diff
@@ -0,0 +1,78 @@
...
+	{
+		$this->startTime = microtime(TRUE);
+		fwrite($this->file, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<testsuites>\n");
+	}
+
+
+	public function result($testName, $result, $message)
+	{
+		$this->buffer .= "\t\t<testcase classname=\"{$testName}\" name=\"{$testName}\"";
+
```
dg: Missing escaping.

Tester/Runner/Output/JUnitWriter.php
```diff
@@ -0,0 +1,78 @@
...
+		if ($result === Runner::FAILED) {
+			$message = str_replace("\n", "\n   ", Dumper::removeColors(trim($message)));
+			$this->buffer .= ">\n\t\t\t<failure message=\"{$message}\"/>\n\t\t";
+		}
```
dg: Escaping

Tester/Runner/Output/JUnitWriter.php
vysinsky: Should I use `htmlspecialchars` or is `addslashes` enough?

Tester/Runner/Output/JUnitWriter.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+
+	public function result($testName, $result, $message)
+	{
+		$testName = '<test"baf">';
+		$this->buffer .= "\t\t<testcase classname=\"" . htmlspecialchars($testName) . "\" name=\"" . htmlspecialchars($testName) . "\"";
+
+		if ($result === Runner::FAILED) {
+			$this->buffer .= ">\n\t\t\t<failure message=\"" . htmlspecialchars($message) . "'\"/>\n\t\t";
```
dg: Why is here apostrof?

Tester/Runner/Output/JUnitWriter.php
```diff
@@ -0,0 +1,78 @@
...
+	public function begin()
+	{
+		$this->startTime = microtime(TRUE);
+		fwrite($this->file, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<testsuites>\n");
+	}
+
+
+	public function result($testName, $result, $message)
+	{
+		$testName = '<test"baf">';
```
dg: Zapomenuty radek

vysinsky: thanks :)

Tester/Runner/CliTester.php
```diff
@@ -184,9 +184,16 @@ private function createRunner()
...
 		if ($this->options['-o'] !== 'none') {
-			$runner->outputHandlers[] = $this->options['-o'] === 'tap'
-				? new Output\TapPrinter($runner)
-				: new Output\ConsolePrinter($runner, $this->options['-s']);
+			switch ($this->options['-o']) {
+				case 'tap':
+					$runner->outputHandlers[] = new Output\TapPrinter($runner);
+					break;
+				case 'junit':
+					$runner->outputHandlers[] = new Output\JUnitWriter($runner);
```
dg: Isn't better to use consistent suffix Printer?

Tester/Runner/Output/JUnitWriter.php
vysinsky: thanks

Tester/Runner/CliTester.php
vysinsky: Yeah, now it does. I used Writer because original purpose of this class was writing xml

Tester/Runner/Output/JUnitPrinter.php
```diff
@@ -0,0 +1,76 @@
...
+			$this->buffer .= ">\n\t\t\t<failure message=\"" . htmlspecialchars($message) . "\"/>\n\t\t";
+		}
+		if ($result === Runner::SKIPPED) {
+			$this->buffer .= ">\n\t\t\t<skipped/>\n\t\t";
+		}
+
+		if ($result === Runner::PASSED) {
+			$this->buffer .= "/>\n";
+		} else {
+			$this->buffer .= "</testcase>\n";
```
dg: IMHO this will produce invalid XML, end of tag is missing. Can you add test?

vysinsky: End of which tag is missing? Yes I will add test.

dg: Maybe more understandable is this

```php
		if ($result === Runner::FAILED) {
			$this->buffer .= ">\n\t\t\t<failure message=\"" . htmlspecialchars($message) . "\"/>\n\t\t</testcase>\n";
		} elseif ($result === Runner::SKIPPED) {
			$this->buffer .= ">\n\t\t\t<skipped/>\n\t\t</testcase>\n";
		} elseif ($result === Runner::PASSED) {
			$this->buffer .= "/>\n";
		}
```

vysinsky: I aggree. I will just use switch instead of if/elseif.

Tester/Runner/CliTester.php
```diff
@@ -184,9 +184,16 @@ private function createRunner()
...
+				case 'tap':
+					$runner->outputHandlers[] = new Output\TapPrinter($runner);
+					break;
+				case 'junit':
+					$runner->outputHandlers[] = new Output\JUnitPrinter($runner);
+					break;
+				default:
+					$runner->outputHandlers[] = new Output\ConsolePrinter($runner, $this->options['-s']);
+			}
 		}
```
milo: @vysinsky Check out the `CliTester::finishCodeCoverage()` echoing.

vysinsky: good point, thanks

Tester/Runner/CliTester.php
```diff
@@ -222,7 +229,7 @@ private function prepareCodeCoverage()
 	/** @return void */
 	private function finishCodeCoverage($file)
 	{
-		if ($this->options['-o'] !== 'none' && $this->options['-o'] !== 'tap') {
+		if ($this->options['-o'] !== 'none' && $this->options['-o'] !== 'tap' && $this->options['-o'] !== 'junit') {
```
JanTvrdik:
~~~php
if (!in_array($this->options['-o'], array('none', 'tap', 'junit'), TRUE) {
~~~

dg:
@milo can I merge it?

tests/Runner/Output/JUnitPrinter.phpt
```diff
@@ -0,0 +1,44 @@
...
+$runner->outputHandlers[] = $printer;
+ob_start();
+$runner->run();
+$output = ob_get_clean();
+
+$expected = <<<XML
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+	<testsuite errors="1" skipped="1" tests="3" time="%a%" timestamp="%a%">
+		<testcase classname="Output/cases/fail.phptx" name="Output/cases/fail.phptx">
```
milo: @vysinsky The paths will fail on Windows. You can use [`%ds%`](https://github.com/nette/tester/blob/f999bdac92db8a35d07905ebdd78d383ad7a22f0/Tester/Framework/Assert.php#L35) not documented "yet".

vysinsky: ok, thanks @milo I am going to fix it

milo:
@dg @vysinsky The tests path `Runner/Output` may collide with dumped `output`. Even no conflicts can occure now, seems better to me to separate code and dumps. Maybe into `tests/RunnerOutput` folder (not sure)?

milo:
Otherwise, I tested it and agree with merging! :+1:

dg:
Flat structure is better, :+1: for `RunnerOutput`

vysinsky:
@milo @dg fixed :)

dg:
Has HHVM problem with error code Job::FAIL? See https://travis-ci.org/nette/tester/jobs/54862342

milo:
> Has HHVM problem with error code Job::FAIL

I'm not familiar with something like that. Must go now, I can test it later if noone before me.

milo:
@dg Looks like HHVM problem #216

Anyway, I think that implementation is OK and :+1: for merge.

milo:
@vysinsky Thank you!

vysinsky:
No problem @milo I really enjoyed my first contribution to Nette :) Thanks for feedback

PR #219, opened 2015-03-26
/Tester changed to /src

dg:


milo:
:+1: Better sooner than later.
:+1: for 2nd commit too. The `tester.php` is mentioned [in documentation](http://tester.nette.org/#toc-running-the-tester). It can be deprecated in next minor and removed in next one.

Btw. you changed the `tester` mode from 755 to 644, so Travis fails.

dg:
fixed

PR #224, opened 2015-04-13
Add numeric to Assert::type

MilanPala:
- feature
- documentation - not needed
- BC break - no

In my opinion `Assert::type` should have `numeric` as one of types.


JanTvrdik:
`is_numeric` is pretty useless for, because it accepts too many things, e. g. `"     -.4e2"` or `" 0xaa"`

MilanPala:
OK, I will try another assertion with better type control.

PR #231, opened 2015-06-10
Fix for the CloverXML statements counting

milo:


milo:
Pfmf, Travis fails because order of `<file>` tags is swapped. I guess the [`RecursiveDirectoryIterator`](https://github.com/nette/tester/blob/a29a3e8fd3084b411c1bc9057460bce8fccd9b4a/src/CodeCoverage/Generators/AbstractGenerator.php#L96) order is somehow changed across the PHP versions. I'll fix it somehow.

PR #236, opened 2015-06-24
DomQuery: fixed css2xpath for brackets in selector

milo:


src/Framework/DomQuery.php
```diff
@@ -81,7 +81,16 @@ public static function css2xpath($css)
...
+					([~*^$]?)=(
+						(?:"[^"]*")|
+						(?:\'[^\']*\')|
+						(?:[^\]]+)
+					)
+				)?
+			\]|                                       # [attr=val] (3,4,5)
 			\s*([>,+~])\s*|                           # > , + ~ (6)
 			(\s+)|                                    # whitespace (7)
 			(\*)                                      # * (8)
```
milo: @dg I noticed that `$m[8]` is not used anywhere. Not sure how are the tests passing :)

src/Framework/DomQuery.php
```diff
@@ -81,7 +81,16 @@ public static function css2xpath($css)
 		$xpath = '//*';
 		preg_match_all('/
 			([#.:]?)([a-z][a-z0-9_-]*)|               # id, class, pseudoclass (1,2)
-			\[([a-z0-9_-]+)(?:([~*^$]?)=([^\]]+))?\]| # [attr=val] (3,4,5)
+			\[
+				([a-z0-9_-]+)
+				(?:
+					([~*^$]?)=(
+						(?:"[^"]*")|
+						(?:\'[^\']*\')|
```
dg: Is (?: needed?

milo: Actully it's not. Would you prefer this multiline version or oneline only?

dg: Multiline is fine. Single line is write-only ;)

milo:
Updated

dg:
Gr8!

milo:
Thanks for review.

PR #237, opened 2015-08-03
Coverage: updated source code template

Mikulas:
- fixes line number alignment (this was already implemented but never worked properly)
- fixes copying code should not also copy line numbers
- updated colors, font and spacing

original:
screen shot 2015-08-03 at 12 04 33

proposed:
screen shot 2015-08-03 at 12 03 15

### DIFF
--- a/src/CodeCoverage/Generators/template.phtml
+++ b/src/CodeCoverage/Generators/template.phtml
@@ -150,27 +150,29 @@
 
 		<div class="file" id="fragment<?php echo $id ?>">
 			<div class="lines">
-				<?php
-					$source = highlight_file($info->file, TRUE);
-					$lineCount = count(explode('<br />', $source));
+				<code>
+					<?php
+						$source = highlight_file($info->file, TRUE);
+						$lineCount = count(explode('<br />', $source));
 
-					$prevColor = NULL;
-					for ($n = 1; $n < $lineCount; ++$n) {
-						if (isset($info->lines[$n]) && isset($classes[$info->lines[$n]])) {
-							$color = $classes[$info->lines[$n]];
-						} else {
-							$color = '';  // not executable
-						}
-						if (!$prevColor) {
-							$prevColor = $color;
-						}
+						$prevColor = NULL;
+						for ($n = 1; $n < $lineCount; ++$n) {
+							if (isset($info->lines[$n]) && isset($classes[$info->lines[$n]])) {
+								$color = $classes[$info->lines[$n]];
+							} else {
+								$color = '';  // not executable
+							}
+							if (!$prevColor) {
+								$prevColor = $color;
+							}
 
-						if ($color) {
-							echo "<div class=\"color $color\"></div>";
+							if ($color) {
+								echo "<div class=\"color $color\"></div>";
+							}
+							echo "$n<br />";
 						}
-						echo "$n<br />";
-					}
-				?>
+					?>
+				</code>
 			</div>
 			<div class="source">
 				<?php echo $source ?>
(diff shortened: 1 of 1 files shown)


fprochazka:
Much nicer!

JanTvrdik:
:+1:

vojtech-dobes:
I think it's not good that it looks like diff, because those are completely different things :).

JanTvrdik:
@vojtech-dobes It looked like diff before, now it just looks like a prettier diff :smile:

vojtech-dobes:
That's true... power of habit :).

dg:
This produces invalid HTML soup:

```html
<tr>
	<td class="line-num">1</td>
	<td class="code">
		<code><span style="color: #06B"><span style="color: #000">&lt;?php
	</td>
</tr>
<tr>
	<td class="line-num">2</td>
	<td class="code"></span></td>
</tr>

```

Mikulas:
Good point, I didn't check what markup `highlight_file` produces.

Instead of parsing `highlight_file` output, I've changed the markup to two side-by-side divs; one for line numbers and the other one for the source code.

If this revision is acceptable, let me know and I'll polish the stylesheet a bit.

milo:
:+1: for some facelift

According to https://validator.w3.org it the HTML still invalid.

According to the following screenshot:
screenshot012

- imho, darker background (than white) + border is better for reading, page orientation and saves eyes
- line highlight is shifted according to code

Mikulas:
Fixed: html validity and mismatched line numbering.

I don't agree the original colors are easier to read. Most important lines are the red ones and since syntax highlighting outputs red, the resulting contrast was terrible. I've intentionally used same color schema github uses: we are very accustomed to this look.

milo:
@Mikulas GitHub uses other colors for highlighting. Would you please try the border and some (I don't know, very light gray) background for code blocks?

The reason is, when I'm reading it on a display with a very high resolution, high contrast and brightness (due to sun light in your back) and I toggle 3 source files to show coverage source, the blocks of code just get fused into one white area.

milo:
Btw, how look lines with dead code?

Mikulas:
Updated with slightly darker colors and border.

- original: https://www.dropbox.com/s/204a41iyxqym56w/0807e28b615af7a51acb6903d9f562ccab15fb51.html?dl=0
- proposed@d6dedaa: https://www.dropbox.com/s/dccdutyi5m8wnh3/d6dedaad97f72e914454881aa8863f7bbfdd2fc0.html?dl=0
- proposed@301fe1e: https://www.dropbox.com/s/unyq6ejh6e3x5c4/301fe1ecc7a3649cf848f5422166095d5738f6d0.html?dl=0

milo:
@Mikulas Much better with a border and IMHO even better with `background: #F8F8F8` for the rest. Btw. lines are still shifted.

jkuchar:
Much better! :+1:

src/CodeCoverage/Generators/template.phtml
```diff
@@ -102,46 +148,37 @@
...
-						} elseif (substr($m, -1) !== '/') {
-							$openTags[] = $m;
+		<div class="file" id="fragment<?php echo $id ?>">
+			<div class="lines">
+				<code>
+					<?php
+						$source = highlight_file($info->file, TRUE);
+						$lineCount = count(explode('<br />', $source));
+
+						$prevColor = NULL;
```
milo: For what it is good now?

milo:
Another style bug.
screenshot013

milo:
@Mikulas Want you to continue on this or may I take over?

Mikulas:
This works for me well enough (well, without rewriting half of code base); feel free to change this however you like.

milo:
Solved by 928cc28e63bb1c10d3e4a78b853758b8c901a7c3

PR #238, opened 2015-08-05
Exceptions from setUp and tearDown are not subject for @throws annotation

MartinMystikJonas:
When Exception is thrown from setUp or TearDown it was processed same as when it is thrown from test itself. Combined with @throws annotation this can lead to false positive tests. For example when exception in expected from test but was not thrown due to error byt this error in test also caused throwing od exception from tearDown it passes.

Also covers case when exception is thrown both from test and tearDown.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -103,10 +103,28 @@ class TestCase
 			try {
 				if ($info['throws']) {
 					$tmp = $this;
-					$e = Assert::error(function () use ($tmp, $method, $args) {
-						$tmp->runTest($method->getName(), $args);
-					}, $throws[0], $throws[1]);
-					if ($e instanceof AssertException) {
+					$e = null;
+					try {
+						$e = Assert::error(function () use ($tmp, $method, $args) {
+							$tmp->runTest($method->getName(), $args);
+						}, $throws[0], $throws[1]);
+						if($e) throw $e;
+					} catch (TestCaseException $e) {
+						if ($e->getPrimaryException() instanceof AssertException) {
+							throw $e->getPrimaryException();
+						}
+						if (isset($throws[0]) && is_string($throws[0]) && preg_match('#^E_[A-Z_]+\z#', $throws[0])) {
+							throw $e;
+						}
+						Assert::error(function () use ($e) {
+							if ($e->getPrimaryException()) { throw $e->getPrimaryException(); };
+						}, $throws[0], $throws[1]);
+						throw $e;
+					} catch (AssertException $e) {
+						throw $e;
+					} catch (\Exception $e) {
+					}
+					if ($e instanceof AssertException || $e instanceof TestCaseException) {
 						throw $e;
 					}
 				} else {
@@ -114,6 +132,8 @@ class TestCase
 				}
 			} catch (AssertException $e) {
 				throw $e->setMessage("$e->origMessage in {$method->getName()}" . (substr(Dumper::toLine($args), 5)));
+			} catch (TestCaseException $e) {
+				throw $e->appendMessage(" in {$method->getName()}()");
 			}
 		}
 	}
@@ -125,7 +145,11 @@ class TestCase
 	 */
 	public function runTest($name, array $args = array())
 	{
-		$this->setUp();
+		try {
+			$this->setUp();
+		} catch (\Exception $setUpEx) {
+			throw new TestCaseException('setUp() phase failed', 0, $setUpEx);
+		}
 		try {
 			call_user_func_array(array($this, $name), $args);
 		} catch (\Exception $e) {
@@ -133,7 +157,7 @@ class TestCase
 		try {
 			$this->tearDown();
 		} catch (\Exception $tearDownEx) {
-			throw isset($e) ? $e : $tearDownEx;
+			throw new TestCaseException('tearDown() phase failed', 0, $tearDownEx, isset($e) ? $e : NULL);
 		}
 		if (isset($e)) {
 			throw $e;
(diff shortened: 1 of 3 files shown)


MartinMystikJonas:
Any idea how to find where exactly code checker failed?

milo:
Related #52.

What about simplier `TestCaseException` with message like `setUp() failed`, `$e` as previous and enhace condition [here](https://github.com/nette/tester/blob/8267b22a84784cfb4427ed09539936b8d84e8e6c/src/Framework/TestCase.php#L109) for `TestCaseException`?

MartinMystikJonas:
#52 is my original solution of similar problem, but it turns out it is not sufficient with combination with @throws annotation

I added condition for TestCaseException and made TestRunException extends TestCaseException. But have it as separate class is important for case when exception both from test itslef and tearDown is thrown - in that case it is important to see both exceptions when you trying to figure what went wrong. Also it is important for this exceptions to carry information about what test method caused this failure otherwise it would be very difficult to tract source of error.

See new version.

milo:
> when exception both from test itslef and tearDown is thrown - in that case it is important to see both exceptions

Why need to see both? Because of exact call order (setUp() -> test() -> tearDown()), you just need to see the first which happend. After that, test exits. You fix it and when thrown in next phase, you fix the next one.

Method name appending is important, but separate it into next commit please.

MartinMystikJonas:
You need to see both in case firt exception (from test) is expected (specified by @throws) but second exception from tearDown is not. For example test correctly throws exception byt Mockery checks in tear down fails then. Without seeing both exceptions you will only know that test correctly throws exception but not what failed then in tearDown.

milo:
Still don't understand :o) What's wrong on [this](https://github.com/milo/tester/commit/bb49ac46cd11eb3340f709f872a0f75b2c235055) solution?

MartinMystikJonas:
It can pass when should not. If test throw SomeException which is expected and tearDown throws MockeryException then MockeryException will be ignored. And if SomeException is expected by @throws annotation test will pass even if some mockery expectations is not met.

milo:
I see now. What about [such](https://github.com/milo/tester/commit/9b2245668baa08af59dc7d9f1861b733eb3189c1) fix?

MartinMystikJonas:
I'm not sure it will work. Can you test it with tests I provided?

But I do not like new state introduced for TestCAse. I think my solution with specialized exception is cleaner.

milo:
It will work. I don't like the exception message mangling in this PR and I prefer the more language natural way.

@dg Don't you have an idea how to solve this?

MartinMystikJonas:
Well what about solution where TestCaseExceotion could contain two original expections.
```
throw new TestCaseException("tearDownFailed", 0, $testException, $tearDownException)
```
And then process this similary to your solution but take secondException from thrown exception instead of TestCAse hidden state?

MartinMystikJonas:
Your solution will not work in two cases:

1) 
Test annotated `@throws MyException` 
It will throw `UnexpectedException`, then tearDown fails. 
Result: you see that TestCAseException is not expected MyException, but information about UnexpectedException which is main cause is lost.

2)
Test annotated `@throws MyException` 
Test by error dont throw anything`, then tearDown fails. 
Result: You see that TestCaseException is not expected MyException, but information that test don't throw anything is lost.

I tried different solutions for almost two hours but I did not find any which can cover all use cases. I think we have only two options:
A) My solution with combined Exceptions to prevent lost of any information
B) Refactor TestCase and move this:
```
$e = Assert::error(function () use ($tmp, $method, $args) {
  $tmp->runTest($method->getName(), $args);
}, $throws[0], $throws[1]);
```
inside runTest method so it wraps only test method itself and not setUp and tearDown to prevent checking tearDown exceptions agains expectations defined by @throws for test method.

MartinMystikJonas:
I was able to create solution which covers most use casses correctly. It is quite complicated but works as expected. See tests.

MartinMystikJonas:
@milo: How to you want to continue on this? I would propose at least do quick fix which will prevent false passing tests ASAP. Then we can discuss what to do next.

milo:
It is nasty bug, but it is present ages here so it will wait. Right now, I'am trying some ways...

MartinMystikJonas:
From my POW best solution will be to apply @thorws annotation checks only to test method itself inside 'runTest'. Question is how to pass needed parameters there. What do you think about third parameter to runTest?

```
public function runTest($name, array $args = array(), $throws = NULL) {
  //...
  $tmp = $this;
  $e = Assert::error(function () use ($tmp, $name, $args) {
    call_user_func_array(array($this, $name), $args);
  }, $throws[0], $throws[1]);
  //...
}
```

milo:
@MartinMystikJonas What about this solution? https://github.com/milo/tester/commit/bb769e67d198c819bcdce3fd1cb7aebf538d4b96

MartinMystikJonas:
Looks good. 

My only objection is using testRunner as TestCase hidden state. It's does not look clean to me. But if we do not want introduce new parameter in runTest it is only way so I think it is ok.

MartinMystikJonas:
Found one problem: When test throws exception tearDown is not called. Common approach is to call tearDown regardless is test suceed. It is because tearDown can contain clean up rutines needed to other test work properly. It prevents cacsade failure triggered by one failing test.

milo:
Good point. Is's obviously not tested :o) I'll append it.

MartinMystikJonas:
Final solution looks good. But I do not think it is BC break. Method runTest does not check expected exception in previous version too. Or am I missing something?

milo:
There is a small probability that someone implemented own `runTest()` and didn't call parent. The expected Exception has been caught in `runMethod()` but now, it will not be catched and test will fail.

MartinMystikJonas:
I see.

PR #239, opened 2015-08-05
Coverage: add support for coversNothing and covers

### LINKED ISSUE #222: Support restricting code coverage
Code coverage results would be a lot more useful if Tester added support for something like PHPUnit's `@covers` and `@coversNothing` annotations ([documentation](https://phpunit.de/manual/current/en/code-coverage-analysis.html#code-coverage-analysis.specifying-covered-methods)).

> It is also possible to specify that a test should not cover any method by using the @coversNothing annotation. This can be helpful when writing integration tests to make sure you only generate code coverage with unit tests. 

This annotation is used to explicitly say what the given test is meant to cover. Coverage of other things is ignored.

Mikulas:
Closes #222, cc @JanTvrdik 

Adds support for
- `@coversNothing`
- `@covers ClassName`
- `@covers ClassName::method`, `@covers ClassName::method()`

Downsides:
- `@covers` only supports FQN

Discussion: These annotations are mostly useful when actually required for all files. There are multiple options of enforcing this:
- require either `@coversNothing` or `@covers` in each test file, which would be a huge bc break
- automatically detect cover annotations, if at least one is present, force them in all tests. Minor bc break, hard to implement.
- introduce a new flag (and maybe switch default behaviour in future releases)

### DIFF
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -60,7 +60,7 @@ class Collector
 	protected static function getCoverFilters()
 	{
 		$annotations = static::getCoverAnnotations();
-		if (isset($annotations['coversNothing'])) {
+		if (isset($annotations['coversnothing'])) {
 			if (isset($annotations['covers'])) {
 				throw new \Exception('Using both @covers and @coversNothing is not supported');
 			}
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -73,7 +73,7 @@ class Collector
 			return self::COVER_ALL;
 		}
 
-		$filters = [];
+		$filters = array();
 		foreach ((array) $annotations['covers'] as $name) {
 			$ref = NULL;
 			try {
@@ -129,7 +129,7 @@ class Collector
 				continue;
 			}
 
-			$refs = isset($filters[$filename]) ? $filters[$filename] : [];
+			$refs = isset($filters[$filename]) ? $filters[$filename] : array();
 			foreach ($lines as $num => $val) {
 				if ($filters === self::COVER_NOTHING || ($filters !== self::COVER_ALL && !static::isCovered($refs, $num))) {
 					$val = -1;
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -7,14 +7,21 @@
 
 namespace Tester\CodeCoverage;
 
+use Tester\Environment;
+use Tester\Helpers;
+
 
 /**
  * Code coverage collector.
  */
 class Collector
 {
+	const COVER_NOTHING = 1;
+	const COVER_ALL = 2;
+
+
 	/** @var resource */
-	private static $file;
+	protected static $file;
 
 
 	/**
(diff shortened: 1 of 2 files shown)


src/CodeCoverage/Collector.php
```diff
@@ -38,6 +45,72 @@ public static function start($file)
...
+		return $filters;
+	}
+
+
+	/**
+	 * @param \ReflectionClass[]|\ReflectionMethod[] $refs
+	 * @param int                                    $line
+	 * @return bool
+	 */
+	private static function isCovered(array $refs, $line)
```
JanTvrdik: I think it would be better to pass `$filename, $filters, $line` and move a bit more logic from `save` here.

~~~php
if ($filters === self::COVER_ALL) {
     return TRUE;
} elseif ($filters === self::COVER_NOTHING || empty($filters[$filename])) {
    return FALSE;
} else {
    foreach ($filters[$filename] as $ref) {
        // ...
    }
}
~~~

Mikulas: I wanted to move as much logic as possible from the inner loop. You added two `$filters[$filename]` lookups that would be invoked for each line.

I'll benchmark it though, it will probably not be any meaningful difference and I'll refactor it then.

milo:
How this behaves with a `@testCase`? In that case, every `testXxx()` method runs in own process. Seems to me the `@covers` has false behaviour.

> - require either @coversNothing or @covers in each test file, which would be a huge bc break

We cannot force people to use it when they use `--coverage`

> - automatically detect cover annotations, if at least one is present, force them in all tests. Minor bc break, hard to implement.

Seems better to me. It's without BC, this is a new feature. It can be done in [init pahse](https://github.com/nette/tester/blob/8267b22a84784cfb4427ed09539936b8d84e8e6c/src/Runner/TestHandler.php#L35).
What about: when `@covers` detected, all not-annotated covers nothing? So `@coversNothing` not needed, but introduce `@coversAll`.

> - introduce a new flag (and maybe switch default behaviour in future releases)

To much complicated for now (and hard to explain in doc :)).

Mikulas:
> How this behaves with a @testCase? In that case, every testXxx() method runs in own process.

Sure, that is not changed, this pull request only filters whatever the process collected. It merges same as it did before.

> Seems to me the @covers has false behaviour.

Please elaborate. I believe the final coverage is correct, the new part is somewhat tested.

> Automatically detect cover annotations … It's without BC

It's slight bcb if an existing test already has a `covers` annotation. I might have been copied from a phpunit project, or just a documentation/reminder for developers what is the class testing.

> What about: when @covers detected, all not-annotated covers nothing? So @coversNothing not needed, but introduce @coversAll

That kind of defeats the purpose. Generally it should be either use `covers` everywhere or not at all, as defaulting to `coversNothing` is very error prone (on the users part).

milo:
> That kind of defeats the purpose. Generally it should be either use covers everywhere or not at all, as defaulting to coversNothing is very error prone (on the users part).

Agree. I wanted to say that `@coversNothing` is not needed. When `@covers` used somewhere, only such reports will be collected.

dg:
Closing, This is too complicated with little benefit.

PR #240, opened 2015-08-05
Dumper: fix php outputs not ending with newline

Mikulas:
This is somewhat irritating when using diff, as some tools warn about missing newlines, making the diff harder to read.

```php
Assert::same([], ['foo']);
```

before:
```
$ diff a b
1c1
< array()
\ No newline at end of file
---
> array('foo')
\ No newline at end of file
```
after
```
$ diff a b
1c1
< array()
---
> array('foo')
```

### DIFF
--- a/src/Framework/Dumper.php
+++ b/src/Framework/Dumper.php
@@ -326,7 +326,7 @@ class Dumper
 			$path = dirname($testFile) . DIRECTORY_SEPARATOR . $path;
 		}
 		@mkdir(dirname($path)); // @ - directory may already exist
-		file_put_contents($path, is_string($content) ? $content : self::toPhp($content));
+		file_put_contents($path, is_string($content) ? $content : (self::toPhp($content) . "\n"));
 		return $path;
 	}
 


src/Framework/Dumper.php
```diff
@@ -326,7 +326,7 @@ public static function saveOutput($testFile, $content, $suffix = '')
 			$path = dirname($testFile) . DIRECTORY_SEPARATOR . $path;
 		}
 		@mkdir(dirname($path)); // @ - directory may already exist
-		file_put_contents($path, is_string($content) ? $content : self::toPhp($content));
+		file_put_contents($path, is_string($content) ? $content : self::toPhp($content) . "\n");
 		return $path;
```
dg: `.` has higher precedence than ternary operator

Mikulas: That's intentional, I didn't want to introduce newline to string `$content`. If that does not end with newline, diff tools should still show the original warning.

milo:
Imho :+1: but if you could add parenthesis...

src/Framework/Dumper.php
fprochazka: @Mikulas please use `()` to make it visually clear what is the intention.

Mikulas:
updated

milo:
@Mikulas Could you change commit message to something like:
```
Dumper: append new line to saved output

Prevents warning 'No newline at end of file' in GNU diffutils.

Mikulas:
updated

milo:
Thank you.

PR #241, opened 2015-08-06
Coverage optimizations

Mikulas:
- Fixed bug that caused (empty) coverage to be generated even for `--method=nette-tester-list-methods`, improving startup time for codebases with a lot of test cases
- <del>Reduced blocking time of coverage to 3 % of original.</del> (merged in https://github.com/nette/tester/commit/9b62462d832f00f881fb84857986bfb51519505f)

Collector benchmark:
```
original blocking: 74.48 ms

     new prepared: 71.04 ms
     new blocking:  2.35 ms
          new sum: 73.38 ms
```

### DIFF
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -7,7 +7,6 @@
 
 namespace Tester\CodeCoverage;
 
-
 /**
  * Code coverage collector.
  */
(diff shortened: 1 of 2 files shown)


src/CodeCoverage/Collector.php
```diff
@@ -45,21 +45,25 @@ public static function start($file)
 	 */
 	public static function save()
 	{
-		flock(self::$file, LOCK_EX);
-		fseek(self::$file, 0);
-		$coverage = @unserialize(stream_get_contents(self::$file)); // @ file may be empty
+		$newCoverage = array_filter(xdebug_get_code_coverage(), 'file_exists', ARRAY_FILTER_USE_KEY);
+
```
dg: ARRAY_FILTER_USE_KEY exists since PHP 5.6

Mikulas: Oh, right, that part is not needed anyway. I'll rebase.

milo:
What about not to set `Environment::COVERAGE` in [initial](https://github.com/nette/tester/blob/fc6d85565097528168ebd9f106744a22d8155ad1/src/Runner/TestHandler.php#L143) phase?

Mikulas:
I've looked into that but it seemed it would require a lot of refactoring, which I deemed unnecessary.

Mikulas:
This was partially incorporated by https://github.com/nette/tester/commit/9b62462d832f00f881fb84857986bfb51519505f .

----

> bug that caused (empty) coverage to be generated even for --method=nette-tester-list-methods, improving startup time for codebases with a lot of test cases

still needs to be fixed though.

milo:
@Mikulas Do you think that something from this PR is needed after bef60c6e58cecee3349d47f7736000e709c37400?

Mikulas:
I believe everything from this PR was merged. Thanks :)

PR #242, opened 2015-08-10
TestCase: do not mask setUp() and tearDown() exceptions

milo:
Currently, when test method is annotated by `@throws` and test method throws an expected exception, the `runTest()` always [escapes by throwing it](https://github.com/nette/tester/blob/fc6d85565097528168ebd9f106744a22d8155ad1/src/Framework/TestCase.php#L136) (or it can be thrown by `setUp()`) and exception from `tearDown()` is never thrown. It can be AssertException and it is buggy to mute it.

Ref #238

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -45,10 +45,10 @@ class TestCase
 
 		if ($method === NULL) {
 			foreach ($methods as $method) {
-				$this->runMethod($method);
+				$this->runTest($method);
 			}
 		} elseif (in_array($method, $methods, TRUE)) {
-			$this->runMethod($method);
+			$this->runTest($method);
 		} else {
 			throw new TestCaseException("Method '$method' does not exist or it is not a testing method.");
 		}
@@ -57,16 +57,17 @@ class TestCase
 
 	/**
 	 * Runs the test method.
+	 * @param  string  test method name
+	 * @param  array  test method parameters (dataprovider bypass)
 	 * @return void
 	 */
-	private function runMethod($method)
+	public function runTest($method, array $args = NULL)
 	{
 		$method = new \ReflectionMethod($this, $method);
 		if (!$method->isPublic()) {
 			throw new TestCaseException("Method {$method->getName()} is not public. Make it public or rename it.");
 		}
 
-		$data = array();
 		$info = Helpers::parseDocComment($method->getDocComment()) + array('dataprovider' => NULL, 'throws' => NULL);
 
 		if ($info['throws'] === '') {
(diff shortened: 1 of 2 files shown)


milo:
@dg May I ask for rewiev?

dg:
What about to process annotation `@throws` inside `runTest()` or pass `$throws` as 3th argument?

Currently you can run `runTest('A')` while `$testRunner` contains "instructructions" for method B.

milo:
That's what @MartinMystikJonas wanted too. I wanted to keep method signature. But if you are OK with it I'll change it.

dg:
Simply, `runTest()` is currently wrong and should be fixed, but not only when is used via `runMethod()`, it should be fixed when is called directly. 

So probably best fix is to move processing of `@throws` in `runTest()`.

MartinMystikJonas:
@dg: See our discussion about this in https://github.com/nette/tester/pull/238#issuecomment-129365171

milo:
@dg What is your vision for the `runTest()` method? I thought, it is isolated from all annotations on purpose. Like:
```php
$test->run('testOne');      # cares about annotations
$test->runTest('testOne');  # does not care

dg:
The question is: *is possible to run test isolated from all annotations*? I thought so, but it looks I was wrong. Test results vary on annotation `@throws`.

MartinMystikJonas:
@milo: Main problem here is that if you ignore @throws annotation it will change test result. I think test result should be same when called as single test through runTest as when called through run.

milo:
@MartinMystikJonas I understand, but I thought that `$test->run('testOne');` is here for such purpose.

@dg Ok, but it changes everything, EVERYTHING! :smiley: I'll update the PR.

milo:
Updated. @dg Not sure what you mean by `move processing of @throws into runTest()`. In this way by 3rd argument? Or parse it again from docblock?

milo:
When I'm thinking about it, what about make `runTest()` deprecated a make `runMethod()` public, add 2nd argumen `$args` (to be able overload dataprovider).

dg:
The „3rd argument“ was just an idea, but wrong. The only way is to parse it again.

milo:
When I'm thinking about it, what about make `runTest()` deprecated and make `runMethod()` public with 2nd argument `$args` (to be able overload dataprovider).

dg:
It is possible too.

MartinMystikJonas:
What was purpose of runTest in oroginal interface? I understand it is so people can extends it but why is this extension needed? 

I would prefer solution with split parse of @throws and @dataProvider annotations. If performance is issue what about create simple helper to get Annotation value for example getTestMethodAnnotation($method, $annotationName) which will parse annotations only once, cache result and then returns only required annotation.

milo:
[Here](https://github.com/nette/tester/blob/7c09d171f4bf232233975311f531faf41c612113/src/Framework/TestCase.php) is the TestCase with deprecated `runTest()`. Maybe next step is to refactor `run($method)`.

dg:
Performance is not issue, but `Helpers::parseDocComment()` can be cached for good feeling :-)

Btw, is really `runMethod()` better name than `runTest()`? Method is more low level than "test".

milo:
@dg @MartinMystikJonas Updated.

MartinMystikJonas:
:+1:

dg:
TestCase.php +54 additions and -53 deletions: seems good :-)

dg:
Just for record: it is BC break in case that method has `@dataprovider` and you call `runTest('method')`.

milo:
:o)
Good point. I'll update commit message.

milo:
I wonder, if someone will find the failing test case.

MartinMystikJonas:
I think it is quite probable. Almost any test which use mocks and @throws at same time is potential problem.

Issue #245, opened 2015-08-13
RFC: TestCase events (simplify modularity)


MartinMystikJonas:
Sometimes we use in our tests some additional tools (Mockista, mockery, database connection, PHP server, memcached, dummy presenter, different container configurations,  ...). Current solution is to create some base TestCases with different combinations of these tools. But it get complicated as more and more tools used. Not mentioning you need to call all of them in tearDown and setUp correctly.

We think about solution inspired by Codeception modules. What we need to create something like that is ability of modules to hook on some TestCAse lifecycle events.

Co we suggest to add to TestCase these events which can be hooked from modules to extends TestCAse behaviour:
- beforeSetUp
- afterSetUp
- beforeTearDown
- afterTearDown
- beforeRunTest
- afterRunTest

Usage for Mockery in some module then can look like:

```
class MockeryModule {

  public static function register(TestCase $testCase) {
    $testCase->afterTearDown[] = function($testCase) {
      Mockery::close();
    }
  }

}
```

Or for database extension:

```
class DatabaseModule {

  public static function register(TestCase $testCase) {
    $testCase->beforeSetUp[] = function($testCase) {
      $this->database->connect();
      $this->database->importData();
    }
    $testCase->afterTearDown[] = function($testCase) {
      $this->database->clearData();
    }
  }

}
```

Concept of modules need to be think throught and is not ready yet. It probably should be addon. But adding this events is first step to be able to do this, it is not BC break and can improve extendability of TestCase.

Proposed solution in PR #246

milo:
I [opened](http://forum.nette.org/en/24028-rfc-testcase-events-for-tester#p161184) discussion.

PR #246, opened 2015-08-13
TestCase: events onBeforeRun, onAfterRun, onBeforeRunTest, onAfterRunTest, onBeforeSetUp, onAfterSetUp, onBeforeTearDown, onAfterTearDown

MartinMystikJonas:
Ability to hook on some TestCase lifecycle events, so TestCase behaviour can be easily extended:
- onBeforeRun (only once)
- onAfterRun (only once)
- onBeforeRunTest (once for each test method, once when using dataProvider)
- onAfterRunTest (once for each test method, once when using dataProvider)
- onBeforeSetUp (once for teach test, once for each dataPrivider item)
- onAfterSetUp (once for teach test, once for each dataPrivider item)
- onBeforeTearDown (once for teach test, once for each dataPrivider item)
- onAfterTearDown (once for teach test, once for each dataPrivider item)

Usage for Mockery then can look like:

```
    $testCase->afterTearDown[] = function($testCase) {
      Mockery::close();
    }
```

Removes need to inherit TestCase ad add Mockery close to tearDown. Removes false passign test when forget to call parent tearDown from test.

Also see RFC #245

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -17,6 +17,44 @@ class TestCase
 	const LIST_METHODS = 'nette-tester-list-methods',
 		METHOD_PATTERN = '#^test[A-Z0-9_]#';
 
+	/** @var array */
+	public $onBeforeSetUp;
+
+	/** @var array */
+	public $onAfterSetUp;
+
+	/** @var array */
+	public $onBeforeTearDown;
+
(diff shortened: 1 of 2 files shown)


dg:
I'm closing PR, it's not a much requested feature.

PR #247, opened 2015-08-14
AbstractGenerator: Introduced exception when no coverage data generated

jkuchar:
Introduced exception when one forget to call \Tester\Environment::setup() in tests and running tester with coverage support.

New exception provides hint how to fix the problem.

### DIFF
--- a/src/CodeCoverage/Generators/AbstractGenerator.php
+++ b/src/CodeCoverage/Generators/AbstractGenerator.php
@@ -37,8 +37,11 @@ abstract class AbstractGenerator
 		if (!is_file($file)) {
 			throw new \Exception("File '$file' is missing.");
 		}
-
-		$this->data = @unserialize(file_get_contents($file)); // @ is escalated to exception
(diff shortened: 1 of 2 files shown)


jkuchar:
BTW this is not ideal. Because when you have generator at least in one file, this exception does not get thrown. Do you have any idea how to check if generator was initialized in **every** test?

fprochazka:
It could be solved easily for tests with TestCase classes, because those are already scanned for test methods. But I don't know how to solve this ideally with "raw" tests.

jkuchar:
I do not thing that it is good idea to use static analysis for checking if this was set. Check for some environment variable (set by Collector; read by Generator) should more precise and easier to implement.

Or we can add list of started tests using https://github.com/nette/tester/blob/master/src/CodeCoverage/Collector.php#L47.

What do you see like the most straightforward way to go?

milo:
Solved by a65fa39e930204842f48bb44f376e8d9583855bd

PR #248, opened 2015-08-29
HtmlGenerator: Not loaded source files was not counted to coverage

jkuchar:
Source files that was not loaded when executing tests was not counted to coverage. This produces false results.

I've added estimator of executable lines. What do you think about this prototype? I'm not sure about duplicating code between CloverXMLGenerator and HtmlGenerator. PhpParser can provide quite accurate data for estimate of lines that can be estimated.

### DIFF
--- a/src/CodeCoverage/Generators/HtmlGenerator.php
+++ b/src/CodeCoverage/Generators/HtmlGenerator.php
@@ -96,6 +96,8 @@ class HtmlGenerator extends AbstractGenerator
 				$coverage = round($covered * 100 / $total);
 				$this->totalSum += $total;
 				$this->coveredSum += $covered;
+			} else {
+				$this->totalSum += count(file($entry, FILE_SKIP_EMPTY_LINES));
 			}
 
 			$light = $total ? $total < 5 : count(file($entry)) < 50;


milo:
Imho, estimation can be more rough. It never will be precise anyway. What about only removing empty lines and count the rest? It will be very pessimistic, but imho OK for this purpose.

jkuchar:
I've done exactly this in first prototype. Removing comment blocks and counting lines gives ~2x times more lines then this filtered version. This gives developer motivation to have at least some test for each file. Because even loading that file will increase coverage. :)

Or what about counting only functions/methods lines count returned by PhpParser?

milo:
PhpParser is overkill here. Imho, this regexp too (it's not 100% correct). Just remove empty lines, it will be super pessimistic, but who cares in this case.

jkuchar:
:+1 (will fix this tomorrow)

jkuchar:
@milo fixed and rebased onto master

src/CodeCoverage/Generators/HtmlGenerator.php
```diff
@@ -110,4 +113,14 @@ private function parse()
 		}
 	}
 
+	private function getNumberOfNonEmptyLines($file)
+	{
+		$lines = 0;
+		foreach(explode("\n", file_get_contents($file)) as $line) {
```
dg: This will not work with `\r\n`. What about `count(file($file, FILE_SKIP_EMPTY_LINES))`

jkuchar: Thanks! I didn't know about that one.

jkuchar:
@dg removed comment

jkuchar:
@milo is there anything more to do with this PR?

milo:
Thank you. I missed the renewed commit.

PR #249, opened 2015-09-03
TestCaseListener for listening on TestCase events

MartinMystikJonas:
Alternative solution #246 for RFC #245.

See https://forum.nette.org/en/24028-rfc-testcase-events-for-tester#p161184

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -17,6 +17,24 @@ class TestCase
 	const LIST_METHODS = 'nette-tester-list-methods',
 		METHOD_PATTERN = '#^test[A-Z0-9_]#';
 
+	/** @var ITestCaseListener[] */
+	private $listeners = array();
+
+	/**
+	 * Add listener for TestCaseEvents
+	 */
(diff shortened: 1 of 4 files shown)


milo:
Thank you for the PR. But imho this is error prone to method name typo.

What is disadvantage of following?
```php
class TestCase
{
    public function onEvent(string $eventName, callable $cb) {}
}

MartinMystikJonas:
What do you mean by error prone to method name typo? All methods are defined in interfaces.

milo:
@MartinMystikJonas Oh, I'm sorry. I mixed up in my head with other things.

MartinMystikJonas:
@milo Shit happens :-) So what is your opinion on this solution?

jkuchar:
Are there any news on this?

milo:
@jkuchar Not from my side :( Didn't find a time to play with it.

JanTvrdik:
I still thinks it is a bad idea.

MartinMystikJonas:
@JanTvrdik Why is it bad idea?

PR #251, opened 2015-09-08
Assert: added noError() assertion [Closes #250]

### LINKED ISSUE #250: Feature request: Add Assert::ok($message) as contrast to Assert::fail($message)
https://forum.nette.org/cs/24204-jak-testovat-ze-neco-nehodi-vyjimku?

milo:


milo:
@dg which way would you prefer? This one or #252?

PR #252, opened 2015-09-12
added Assert::noError

richard-ejem:
…a piece of code does not fail with error or exception

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -374,6 +374,37 @@ class Assert
 		}
 	}
 
+	/**
+	 * Checks if the function does neither generate PHP error nor throw Exception
+	 * @param  callable
+	 * @param  string
+	 * @return null|\Exception
+	 */
+	public static function noError($function, $message = NULL)
+	{
+		self::$counter++;
+
+		set_error_handler(function ($severity, $msg, $file, $line) use (& $expected, $message) {
+			if (($severity & error_reporting()) !== $severity) {
+				return;
+			}
+			$errorStr = Helpers::errorTypeToString($severity) . ($msg ? " ($msg)" : '');
+			restore_error_handler();
+			Assert::fail(($message ? "$message: " : '') . "occurred error $errorStr");
+		});
+		try {
+			call_user_func($function);
+		} catch (\Exception $e) {
+			if ($e instanceof AssertException) {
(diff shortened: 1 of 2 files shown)


milo:
Try your solution with:
```php
function aaa()
{
	try {
		bbb();
	} catch (\RuntimeException $e) {
		throw new \RuntimeException('Backend failed', 0, $e);
	}
}

function bbb()
{
	throw new \RuntimeException('This is wrong');
}

Assert::noError(function () {
	aaa();
});
```

You get:
```
Failed: RuntimeException was thrown. Code: 0 Message: Backend failed

in tests\Framework\Assert.noError.phpt(46) Tester\Assert::noError()
```

I'm trying to say, that stack trace and well formatted exceptions chain is crucial here. Imho #251 is good enough.

richard-ejem:
Well, I can add support for printing out stack trace, or maybe even semantically better - list of exception tested not to be thrown. i.e. Assert::noError(function() { new X(new Foo)); }, 'Nette\InvalidArgumentException', "X should accept Foo"); - testing if X constructor accepts Foo. if InvalidArgumentException is thrown, shows user-friendly error message. If something other fails, shows stack trace (the assertion just throws the exception out).

If we decide for your super simple solution, I suggest even simpler:

    public static function noError($function) {
        self::$counter++;
        $function();
    }

as calling to self::error($fn, []) doesn't seem anyhow helpful to me.

milo:
The `noError()` assertion cannot have any argument. It asserts that there will be no error. It even cannot have expectation message, you don't know where error comes from.

The `self::error($fn, [])` solution dumps diff into `output` folder with all errors.

> doesn't seem anyhow helpful to me

And what is? I thought that you want to stop use `Assert::true(TRUE)`?

richard-ejem:
I want to quickly see what failed from the tester output and do it in semantically clear and straightforward way.

> you don't know where error comes from

I kindof do. As written in my example before, i.e. I test that a method accepts certain argument, therefore I explicitly test that the metod does not throw InvalidArgumentException.
If it does, I want to just see "should accept SomeTypeOfArgument". If it throws anything other, yes, stack trace is then useful.

milo:
Closed in favour of e9e1cd82d879574bfc93ad19c6bac53124a0f478.

PR #255, opened 2015-09-16
run tearDown() if TestCase fails with PHP error [closes #254]

### LINKED ISSUE #254: tearDown not executed when test fails with error or exception
I suppose this is a bug - this tidy method should be executed no matter if the test succeeds or fails. Actually tearDown works if the test fails on assertion, but not on error/exception.

//Edit: it works on exception, but not on any runtime error.

Please confirm it is a bug, I can fix and pullrequest it.

richard-ejem:


src/Framework/TestCase.php
```diff
@@ -107,8 +107,12 @@ public function runTest($method, array $args = NULL)
 
 		foreach ($data as $params) {
 			try {
+				$me = $this; // PHP 5.3 compatibility
+				$lastHandler = set_error_handler(function() use ($me, &$lastHandler) {
+					$me->failsafeTearDown();
```
milo: Restore handler here to prevent recursion.

src/Framework/TestCase.php
```diff
@@ -174,6 +179,10 @@ protected function tearDown()
 	{
 	}
 
+	/** @internal */
+	public function failsafeTearDown() {
```
milo: Don't add such method for now.

src/Framework/TestCase.php
```diff
@@ -174,6 +179,10 @@ protected function tearDown()
...
+		$this->tearDown();
```
milo: What about error in `tearDown()`. How to handle it?

src/Framework/TestCase.php
```diff
@@ -107,8 +107,12 @@ public function runTest($method, array $args = NULL)
 
 		foreach ($data as $params) {
 			try {
+				$me = $this; // PHP 5.3 compatibility
+				$lastHandler = set_error_handler(function() use ($me, &$lastHandler) {
+					$me->failsafeTearDown();
+					return ($lastHandler === NULL) ? FALSE : $lastHandler();
```
milo: Missing error handler arguments.

src/Framework/TestCase.php
richard-ejem: @milo well, if tearDown() fails, there is nothing more we can do (point of this pull request is to execute tearDown if setUp or test* fails).

src/Framework/TestCase.php
richard-ejem: The only point of this method is to make the stuff PHP 5.3 compatible - make possible to call tearDown from a callback, as PHP 5.3 doesn't see "$this" context from anonymous function.

src/Framework/TestCase.php
richard-ejem: thanks, fixed now

src/Framework/TestCase.php
richard-ejem: done ;)

src/Framework/TestCase.php
```diff
@@ -174,6 +180,10 @@ protected function tearDown()
 	{
 	}
 
+	/** @internal */
+	public function failsafeTearDown() {
```
milo: Name it as `_tearDown()` and fix coding style. Double empty line between methods, one empty line after the last one, curly bracket on separated line.

src/Framework/TestCase.php
```diff
@@ -107,8 +107,13 @@ public function runTest($method, array $args = NULL)
 
 		foreach ($data as $params) {
 			try {
+				$me = $this; // PHP 5.3 compatibility
+				$lastHandler = set_error_handler(function() use ($me, &$lastHandler) {
+					restore_error_handler();
+					$me->failsafeTearDown();
+					return ($lastHandler === NULL) ? FALSE : call_user_func_array($lastHandler, func_get_args());
+				});
 				$this->setUp();
```
milo: Move `setUp()` call before error handler, exceptions aren't catched too.

src/Framework/TestCase.php
```diff
@@ -107,8 +107,13 @@ public function runTest($method, array $args = NULL)
 
 		foreach ($data as $params) {
 			try {
+				$me = $this; // PHP 5.3 compatibility
+				$lastHandler = set_error_handler(function() use ($me, &$lastHandler) {
+					restore_error_handler();
+					$me->failsafeTearDown();
+					return ($lastHandler === NULL) ? FALSE : call_user_func_array($lastHandler, func_get_args());
+				});
```
milo: Fix coding style. Space after `()`, space after `&`.

milo:
What about
```php
	/** @internal */
	public function _errorHandler()
	{
		restore_error_handler();
		@$this->tearDown(); // ignore, error from testMethod() has higher precedence

		return $this->previousHandler
			? call_user_func_array($this->previousHandler, func_get_args())
			: FALSE;
	}
```

src/Framework/TestCase.php
pavelkouril: Also, shouldn't it be `function ()`?

richard-ejem:
@milo thanks for suggestion, I like your _errorHandler alternative :) fixed in my code

milo:
Solved by d25b67216a30a723db040ebcfd2c10e446f12218.

PR #257, opened 2015-09-18
Testcase does not run tearDown() when test fails

milo:
Ref #254

@dg What do you think about this approach?


MartinMystikJonas:
See my objections here: https://github.com/nette/tester/issues/254#issuecomment-141408679

hrach:
:-1:

PR #258, opened 2015-10-02
Support for phpdbg with fast code coverage

JanTvrdik:
Most of critical phpdbg bugs are fixed since RC4, some bugs are however still present.

https://bugs.php.net/bug.php?id=70532
https://bugs.php.net/bug.php?id=70531
https://bugs.php.net/bug.php?id=70614

### DIFF
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -24,41 +24,90 @@ class Collector
 	 */
 	public static function start($file)
 	{
-		if (!extension_loaded('xdebug')) {
-			throw new \Exception('Code coverage functionality requires Xdebug extension.');
-		} elseif (self::$file) {
+		if (self::$file) {
 			throw new \LogicException('Code coverage collector has been already started.');
 		}
-
 		self::$file = fopen($file, 'a+');
-		xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE);
-		register_shutdown_function(function () {
-			register_shutdown_function(array(__CLASS__, 'save'));
+
+		if (defined('PHPDBG_VERSION') && PHP_VERSION_ID >= 70000) {
+			phpdbg_start_oplog();
+			$collector = 'collectPhpDbg';
+
+		} elseif (extension_loaded('xdebug')) {
+			xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE);
+			$collector = 'collectXdebug';
+
+		} else {
+			$alternative = PHP_VERSION_ID >= 70000 ? ' or phpdbg SAPI' : '';
+			throw new \Exception("Code coverage functionality requires Xdebug extension$alternative.");
+		}
+
+		register_shutdown_function(function () use ($collector) {
+			register_shutdown_function(function () use ($collector) {
+				list($positive, $negative) = call_user_func(array(__CLASS__, $collector));
+				self::save($positive, $negative);
+			});
 		});
+	}
+
+
+	/**
+	 * Collects information about code coverage.
+	 * @return array
+	 */
+	private static function collectXdebug()
+	{
+		$positive = $negative = array();
+
+		foreach (xdebug_get_code_coverage() as $file => $lines) {
+			if (!file_exists($file)) {
+				continue;
+			}
+
+			foreach ($lines as $num => $val) {
+				if ($val > 0) {
+					$positive[$file][$num] = $val;
+				} else {
+					$negative[$file][$num] = $val;
+				}
+			}
+		}
+
+		return array($positive, $negative);
+	}
+
+
+	/**
+	 * Collects information about code coverage.
+	 * @return array
+	 */
+	private static function collectPhpDbg()
+	{
+		$positive = phpdbg_end_oplog();
+		$negative = phpdbg_get_executable();
+
+		foreach ($positive as $file => & $lines) {
+			$lines = array_fill_keys(array_keys($lines), 1);
+		}
+
+		foreach ($negative as $file => & $lines) {
+			$lines = array_fill_keys(array_keys($lines), -1);
+		}
+
+		return array($positive, $negative);
 	}
 
 
 	/**
 	 * Saves information about code coverage. Do not call directly.
 	 * @return void
-	 * @internal
 	 */
-	public static function save()
+	private static function save(array $positive, array $negative)
 	{
 		flock(self::$file, LOCK_EX);
 		fseek(self::$file, 0);
-		$coverage = @unserialize(stream_get_contents(self::$file)); // @ file may be empty
-
-		foreach (xdebug_get_code_coverage() as $filename => $lines) {
-			if (!file_exists($filename)) {
-				continue;
-			}
-			foreach ($lines as $num => $val) {
-				if (empty($coverage[$filename][$num]) || $val > 0) {
-					$coverage[$filename][$num] = $val; // -1 => untested; -2 => dead code
-				}
-			}
-		}
+		$original = @unserialize(stream_get_contents(self::$file)) ?: array(); // @ file may be empty
+		$coverage = array_replace_recursive($negative, $original, $positive);
 
 		ftruncate(self::$file, 0);
 		fwrite(self::$file, serialize($coverage));
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -62,7 +62,7 @@ class Environment
 	{
 		self::$useColors = getenv(self::COLORS) !== FALSE
 			? (bool) getenv(self::COLORS)
-			: (PHP_SAPI === 'cli' && ((function_exists('posix_isatty') && posix_isatty(STDOUT))
+			: ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && ((function_exists('posix_isatty') && posix_isatty(STDOUT))
 				|| getenv('ConEmuANSI') === 'ON' || getenv('ANSICON') !== FALSE));
 
 		$colors = & self::$useColors;
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -231,7 +231,8 @@ XX
 	private function prepareCodeCoverage()
 	{
 		if (!$this->interpreter->hasXdebug()) {
-			throw new \Exception("Code coverage functionality requires Xdebug extension (used {$this->interpreter->getCommandLine()})");
+			$alternative = PHP_VERSION_ID >= 70000 ? ' or phpdbg SAPI' : '';
+			throw new \Exception("Code coverage functionality requires Xdebug extension$alternative (used {$this->interpreter->getCommandLine()})");
 		}
 		file_put_contents($this->options['--coverage'], '');
 		$file = realpath($this->options['--coverage']);
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -177,6 +177,8 @@ XX
 
 		if (preg_match('#HipHop VM#', $output)) {
 			$this->interpreter = new HhvmPhpInterpreter($this->options['-p'], $args);
+		} elseif (strpos($output, 'phpdbg') !== FALSE) {
+			$this->interpreter = new ZendPhpDbgInterpreter($this->options['-p'], $args);
 		} else {
 			$this->interpreter = new ZendPhpInterpreter($this->options['-p'], $args);
 		}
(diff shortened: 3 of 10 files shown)


Mikulas:
:trophy:

(also it partially solves #241)

src/CodeCoverage/Collector.php
```diff
@@ -20,20 +25,66 @@ class Collector
...
+		}
+	}
+
+
+	/**
+	 * @return void
+	 */
+	private static function startXdebug()
+	{
+		if (!extension_loaded('xdebug')) {
```
Majkl578: This should (imho) also check `PHP_SAPI !== 'phpdbg'`?

src/Runner/ZendPhpDbgInterpreter.php
```diff
@@ -0,0 +1,102 @@
...
+		$output = stream_get_contents($pipes[1]);
+
+		$this->error = trim(stream_get_contents($pipes[2]));
+		if (proc_close($proc)) {
+			throw new \Exception("Unable to run '$path': " . preg_replace('#[\r\n ]+#', ' ', $this->error));
+		} elseif (!preg_match('#^PHP ([\w.]+)#im', $output, $matches)) {
+			throw new \Exception("Unable to detect PHP version (output: $output).");
+		} elseif (version_compare($matches[1], '7.0.0', '<')) {
+			throw new \Exception('Unable to use phpdbg on PHP < 7.0.0.');
+		}
```
Majkl578: I've seen some really strange behavior when I ran phpdbg with XDebug loaded as a module. Maybe also check for XDebug not being loaded?

Majkl578:
:+1: 

Since phpdbg seems to be much faster than XDebug (and it is also a native part of PHP), how about making it a default interpreter for PHP 7.0+?

---

If anyone is testing this, be aware that before PHP 7.0RC5, there is a bug in phpdbg coverage report when return types are used: https://github.com/php/php-src/commit/6c61286da4f2d71e659da70842168d71835292f9.

src/Runner/ZendPhpDbgInterpreter.php
JanTvrdik: What? Phpdbg is only usable on php7 and xdebux is only usable on php5.

src/CodeCoverage/Collector.php
JanTvrdik: Why? Phpdbg is pointless when you use xdebux strategy. Did you meant to use or or and?

src/Runner/ZendPhpDbgInterpreter.php
Majkl578: XDebug will be usable on PHP 7 as well, once it's ported. It is already working for debugging (not coverage though).

milo:
:+1:

Some Tester's tests are failing for me, I thing we should wait for final 7.0.0. In meantime:
- is the `ZendPhpDbgInterpreter` needed? The only difference is the `-qrr` arg
- I'm thinking about `PhpInterpreter::hasXdebug()` renaming, it's needed for coverage only
- is the `Coverage::STRATEGY_AUTO` needed? Seems better to me send the strategy explicitly somehow from runner

JanTvrdik:
> Some Tester's tests are failing for me

You need RC4 to be able to run phpdbg and RC5 to make all tests OK.

> The only difference is the -qrr arg

No. clipboard01


> I'm thinking about PhpInterpreter::hasXdebug() renaming

Or removing entirely. It's currently not used anywhere.

> Seems better to me send the strategy explicitly somehow from runner

Runner may use different interpreter than jobs.

It's annoying to generate coverage now, because you need to write two long arguments (coverage + coverage-src). Feel free to add coverarage-strategy but the parameter must be optional. The autodetection is simple a will work OK for 99.9 % of people.

JanTvrdik:
Thinking about it – it will probably be even better to split `ZendPhpInterpreter` to `ZendPhpCgiInterpreter` and `ZendPhpCliInterpreter`.

milo:
>> Some Tester's tests are failing for me

> You need RC4 to be able to run phpdbg and RC5 to make all tests OK.

Oh, I cloned from GitHub, there is no RC5 yet.

>> The only difference is the -qrr arg

> No. 

Hmm, `--version` works for both.


>>  Seems better to me send the strategy explicitly somehow from runner

> Runner may use different interpreter than jobs.

Can be detected in an `/info.php` way.

Majkl578:
7.0.0 RC5 is out.

hrach:
:+1: Is there any drawback/issue that needs to be solved?

milo:
Waiting for stable 7 at least.

JanTvrdik:
If anybody is willing to help – download the PR and run tests against latest PHP 7, last time I checked this one test was failing. Unfortunately I did not have the time to determine whether it is bug in tester or PHP 7.

hrach:
> Waiting for stable 7 at least.

Well, I do not want to push on merging a feature with failing tests, however, I see no reason not to merge a feature for an unstable version of php into development branch. Nette is full of support for php7.

Majkl578:
> however, I see no reason not to merge a feature for an unstable version of php into development branch

Definitely :+1:, that argument makes no sense to me.

Btw. PHP 7 release is postponed at least until the end of november.

milo:
RC5 was not failling, later is failing. I don't want to invest my time into bug hunting. Its not only about Tester code, but compiling next RC and setup environment... that's time consuming.

That's the reason why I'm waiting for stable release.

JanTvrdik:
> I don't want to invest my time into bug hunting.

Ehm… now you can fix phpdbg before 7.0.0 released. After release you are fucked.

> why I'm waiting for stable release

It will not start working unless someone will make it work. Waiting is pointless. The situation will only get worse.

milo:
@JanTvrdik Please, delete the vulgar words.

Anyway, if you promise me that you will care about phpdbg and you will solve reported bugs about it, I'll invest the time into it.

hrach:
I didn't want to bring emotions here. Still, the "at least" seems to be restrictive and impropper. :) That's all.

Majkl578:
@Milo: Although I understand your approach which it makes sense for a library, unfortunately it's not that logical for a tool that is used to run tests. Especially when talking about php-next, which is pretty critical now as the GA is around the corner.

And in meantime, PHPUnit has PHP 7 support (incl. phpdbg) for over a month now...

JanTvrdik:
@milo I can't promise that all bugs will be resolved before 7.0.0. I'm saying that they might be fixed. After the release you would need to learn how to time travel.

> I understand your approach which it makes sense for a library

I don't. If you want your library to support PHP 7 you should always start working on that before PHP 7 is released not after. This is especially important for libraries which use low-level or uncommon APIs (such as Tester or Tracy).

src/CodeCoverage/Collector.php
```diff
@@ -13,6 +13,11 @@
  */
 class Collector
 {
+	const
```
dg: What about two Collector classes?

JanTvrdik: There is some common code, so you would need three classes – one for each strategy and one base class. The logic for deciding which strategy to use would probably end-up extracted outside of those classes in \Tester\Environment::setupCoverage.

To summarize – it would require more code but should be possible.

Mikulas:
Ok so the tests fail on 7.0.0 as well.

`Job::getErrorOutput()` will presumably never work, phpdbg send (correctly) all output to stdout, even when in underlaying process prints to stderr.

Proposal A: skip `getErrorOutput` test for phpdbg.
Proposal B: remove `getErrorOutput` altogether from Job and return complete output in `getOutput`. Extensions of `Job` may implement another interface with `getStandardOutput` and `getErrorOutput`

Ansi codes in output could be fixed with this patch
```diff
src/Runner/ZendPhpDbgInterpreter.php:41
+ fwrite($pipes[0], "set colors 0\n");
+
```

Mikulas:
Stderr: @JanTvrdik found it's actually fixed on php-7.0.1RC1 https://github.com/php/php-src/commit/c1189eccae3c555ac5a8610110bb3edb80998d2c

Ansi codes: alternative fix for noninteractive mode which is what Job invokes
```diff
src/Runner/ZendPhpDbgInterpreter.php:63
-		return $this->path . ' -qrr ' . $this->arguments;
+		return $this->path . ' -qrrb ' . $this->arguments;
```

Other than those bugs, are there any issues that would prevent getting this merged?

JanTvrdik:
PING

milo:
@JanTvrdik Thank you for update. I want to review, merge and test it locally before master merge, but I'm quite busy last days and I'll be next week. I'll do it when possible.

JanTvrdik:
The last commit is a bit controversial however I feel like fixing every single library out there to change `PHP_SAPI === 'cli'` to `PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg'` is impossible.

hrach:
@JanTvrdik :+1: :100:

milo:
Ref https://github.com/krakjoe/phpdbg/issues/149

milo:
Ref #279

milo:
@JanTvrdik Would you squash last commit into the 1st and fix Job test in [this way](https://github.com/milo/tester/blob/pull-phpdbg/tests/Runner/Job.phpt#L22)?

milo:
Ping @JanTvrdik

dg:
What about this? https://github.com/dg/tester/commit/330ef182dcdc2f2a36081c9e850144befb8849d4

Majkl578:
Just a note: you can be running PHPDBG SAPI while having XDebug enabled as well, but in that case XDebug is not what we want to use, right?

milo:
@dg Seems better.
@JanTvrdik Some attitude to this?

hrach:
@milo why it seems better? It seems as a spagtheti code, badly testable, wired, ...

dg:
It is only simplified, with fixed exception message.

JanTvrdik:
* yes, there should be check for PHP 7.0.0+; phpdbg in PHP 5.6 does not support code coverage
* with @dg's refactoring it would be difficult to implement command line switch to manually force specific code coverage strategy, so it really depends on whether we see `--coverage-strategy xdebug` as useful or not

milo:
I don't see the reason for choosing the coverage strategy manually for now and I would wait for real-world issue. In that case, David's simplification seems OK to me.

@JanTvrdik Would you add the version check and such simplifications?

JanTvrdik:
@milo done

src/Runner/CliTester.php
```diff
@@ -228,9 +230,6 @@ private function createRunner()
 	/** @return string */
 	private function prepareCodeCoverage()
 	{
-		if (!$this->interpreter->hasXdebug()) {
-			throw new \Exception("Code coverage functionality requires Xdebug extension (used {$this->interpreter->getCommandLine()})");
-		}
```
milo: Imho, this should stay there with modified message. And as a workaround, `ZendPhpDbgInterpreter::hasXDebug()` should return TRUE.

milo:
@JanTvrdik One note in code. I'm sorry I didn't noticed earlier.

src/Runner/CliTester.php
Majkl578: That doesn't make any sense. PHPDBG may or may not XDebug loaded, just like normal CLI/CGI SAPI.

milo: The `ZendPhpDbgInterpreter::hasXDebug()` is used only for detection that Coverage is possible. I have a plan for refactoring.

JanTvrdik: Yes, this should be refactored, but it is outside of the scope of this PR.

JanTvrdik:
Everything related to `hasXDebug` should be refactored, but it is outside of the scope of this PR.

milo:
Agree. I thought only to return TRUE, nothing more.

src/Runner/CliTester.php
milo: @JanTvrdik Or maybe better `... && !$this->interpreter instanceof ZendPhpDbgInterpreter`.

milo:
Thank you!

src/Framework/Environment.php
```diff
@@ -62,7 +62,7 @@ public static function setupColors()
 	{
 		self::$useColors = getenv(self::COLORS) !== FALSE
 			? (bool) getenv(self::COLORS)
-			: (PHP_SAPI === 'cli' && ((function_exists('posix_isatty') && posix_isatty(STDOUT))
+			: ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && ((function_exists('posix_isatty') && posix_isatty(STDOUT))
```
dg: Is `PHP_SAPI === 'phpdbg'` correct?  PHP_SAPI is 'cli' when argument `-qrrb -S cli` is used.

milo: @dg Handy when called `phpdbg path/to/test.phpt` manually.

fprochazka:
Awesome work guys!

PR #259, opened 2015-10-02
Minimal PHP version is 5.4.0

milo:
Maybe 5.4.6 (see bugs sanitization in tests) or even higher?
If agreed, I'll squash it all under the 1st commit.

What about to release 1.5.1 with all commits until Oct 1, 2015 and open the 2.0 version?

### DIFF
--- a/src/Framework/FileMock.php
+++ b/src/Framework/FileMock.php
@@ -22,7 +22,13 @@ class FileMock
 	private $content;
 
 	/** @var int */
-	private $pos;
+	private $readingPos;
+
+	/** @var int */
+	private $writingPos;
+
+	/** @var bool */
+	private $appendMode;
 
 	/** @var bool */
 	private $isReadable;
@@ -74,8 +80,9 @@ class FileMock
 		}
 
 		$this->content = & self::$files[$path];
-		$this->pos = $m[1] === 'a' ? strlen($this->content) : 0;
-
+		$this->appendMode = $m[1] === 'a';
+		$this->readingPos = 0;
+		$this->writingPos = $this->appendMode ? strlen($this->content) : 0;
 		$this->isReadable = isset($m[2]) || $m[1] === 'r';
 		$this->isWritable = isset($m[2]) || $m[1] !== 'r';
 
@@ -89,8 +96,9 @@ class FileMock
 			return '';
 		}
 
-		$res = substr($this->content, $this->pos, $len);
-		$this->pos += strlen($res);
+		$res = substr($this->content, $this->readingPos, $len);
+		$this->readingPos += strlen($res);
+		$this->writingPos += $this->appendMode ? 0 : strlen($res);
 		return $res;
 	}
 
@@ -101,36 +109,41 @@ class FileMock
 			return 0;
 		}
 
-		$this->content = substr($this->content, 0, $this->pos)
-			. str_repeat("\x00", max(0, $this->pos - strlen($this->content)))
+		$length = strlen($data);
+		$this->content = substr($this->content, 0, $this->writingPos)
+			. str_repeat("\x00", max(0, $this->writingPos - strlen($this->content)))
 			. $data
-			. substr($this->content, $this->pos + strlen($data));
-		$this->pos += strlen($data);
-		return strlen($data);
+			. substr($this->content, $this->writingPos + $length);
+
+		$this->readingPos += $length;
+		$this->writingPos += $length;
+
+		return $length;
 	}
 
 
 	public function stream_tell()
 	{
-		return $this->pos;
+		return $this->readingPos;
 	}
 
 
 	public function stream_eof()
 	{
-		return $this->pos >= strlen($this->content);
+		return $this->readingPos >= strlen($this->content);
 	}
 
 
 	public function stream_seek($offset, $whence)
 	{
 		if ($whence === SEEK_CUR) {
-			$offset += $this->pos;
+			$offset += $this->readingPos;
 		} elseif ($whence === SEEK_END) {
 			$offset += strlen($this->content);
 		}
 		if ($offset >= 0) {
-			$this->pos = $offset;
+			$this->readingPos = $offset;
+			$this->writingPos = $this->appendMode ? $this->writingPos : $offset;
 			return TRUE;
 		} else {
 			return FALSE;
@@ -140,8 +153,13 @@ class FileMock
 
 	public function stream_truncate($size)
 	{
+		if (!$this->isWritable) {
+			return FALSE;
+		}
+
 		$this->content = (string) substr($this->content, 0, $size)
 			. str_repeat("\x00", max(0, $size - strlen($this->content)));
+		$this->writingPos = $this->appendMode ? 0 : $this->writingPos;
 		return TRUE;
 	}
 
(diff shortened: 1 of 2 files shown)


Majkl578:
Hmm, is there a reason for keeping compatibility with unsupported PHP 5.4 & 5.5?
(I.e. PHPUnit 5 - released today - requires 5.6 & 7.0 only.)

milo:
I have no idea, how is the 5.6 spreaded. Tester itself does not need 5.6 features necessarily.

Personaly, I'm ok with 5.6 for my projects.

dg:
I would merge it as late as possible.

dg:
Nette 2.3 will be actively maintained to february 2016. After that I will have no project developed under PHP 5.3, so I will not need Tester for 5.3. It does not mean that it is necessary to wait a month, it can be merged sooner.

dg:
If you develop a library for PHP 5.4.0, the testing tool should not limit you and allow you to test it in PHP 5.4.0, so I think that minimal required version should be 5.*.0. And IMHO 5.5/5.6 has too few benefits over 5.4, so minimal required version 5.4.0 seems best to me.

src/CodeCoverage/Generators/template.phtml
```diff
@@ -146,7 +146,7 @@
 	<?php endforeach ?>
 
 	<footer>
-		<p>Generated by <a href="https://tester.nette.org">Nette Tester</a> at <?php echo @date('Y/m/d H:i:s') // @ timezone may not be set ?></p>
+		<p>Generated by <a href="http://tester.nette.org">Nette Tester</a> at <?= @date('Y/m/d H:i:s') // @ timezone may not be set ?></p>
```
dg: Here should stay https

src/Framework/Assert.php
```diff
@@ -283,7 +283,7 @@ public static function type($type, $value, $description = NULL)
 	 * @param  int code
 	 * @return \Exception
 	 */
-	public static function exception($function, $class, $message = NULL, $code = NULL)
+	public static function exception(callable $function, $class, $message = NULL, $code = NULL)
```
dg: … and method Assert::throws()

tests/Runner/Runner.edge.phpt
```diff
@@ -35,7 +31,7 @@ $cli = PHP_SAPI === 'cli';
 $bug62725 = $cli && PHP_VERSION_ID >= 50400 && PHP_VERSION_ID <= 50406;
 Assert::same($bug62725 ? [Runner::PASSED, NULL] : [Runner::FAILED, 'Exited with error code 231 (expected 0)'], $logger->results['shutdown.exitCode.a.phptx']);
 
-$bug65275 = !defined('HHVM_VERSION') && $cli && PHP_VERSION_ID >= 50300; // bug still exists
+$bug65275 = !defined('HHVM_VERSION') && $cli;
 Assert::same($bug65275 ? [Runner::FAILED, 'Exited with error code 231 (expected 0)'] : [Runner::PASSED, NULL], $logger->results['shutdown.exitCode.b.phptx']);
```
dg: Are you sure here?

src/Framework/Dumper.php
```diff
@@ -307,7 +307,7 @@ public static function dumpException($e)
 			. (isset($stored) ? 'diff ' . Helpers::escapeArg($stored[0]) . ' ' . Helpers::escapeArg($stored[1]) . "\n\n" : '');
 
 		foreach ($trace as $item) {
-			$item += array('file' => NULL, 'class' => NULL, 'type' => NULL, 'function' => NULL);
+			$item += ['file' => NULL, 'class' => NULL, 'type' => NULL, 'function' => NULL];
 			if ($e instanceof AssertException && $item['file'] === __DIR__ . DIRECTORY_SEPARATOR . 'Assert.php') {
 				continue;
 			}
```
dg: The line bellow can be simplified to `&& strpos($tmp = file($item['file'])[$item['line'] - 1], "::$item[function](") ? $tmp : NULL;`

dg:
> If agreed, I'll squash it all under the 1st commit.

IMHO more than one commit is better, I used one commit on other repositories and it was not good.

tests/Runner/Runner.edge.phpt
milo: I cannot figure how the removed part could ever be FALSE with PHP 5.3+.

src/Framework/Dumper.php
milo: Thanks, that is I couldn't recall for :)

tests/Runner/Runner.edge.phpt
dg: You right

src/Framework/Dumper.php
dg: This is the most cryptic part of Tester :-)

milo:
>> If agreed, I'll squash it all under the 1st commit.

> IMHO more than one commit is better, I used one commit on other repositories and it was not good.

OK. Plan is rebase it, update it, maybe squash some commits together.

src/Framework/Environment.php
```diff
@@ -70,10 +70,9 @@ public static function setupColors()
 				&& ((function_exists('posix_isatty') && posix_isatty(STDOUT))
 					|| getenv('ConEmuANSI') === 'ON' || getenv('ANSICON') !== FALSE) || getenv('term') === 'xterm-256color');
 
-		$colors = & self::$useColors;
-		ob_start(function ($s) use (& $colors) {
-			return $colors ? $s : Dumper::removeColors($s);
-		}, PHP_VERSION_ID < 50400 ? 2 : 1, FALSE);
+		ob_start(function ($s) {
```
dg: This should probably stay https://github.com/nette/tester/commit/d76cd9776f757763ce1479602c1c3232c95c0a81

src/CodeCoverage/Generators/AbstractGenerator.php
```diff
@@ -96,31 +96,13 @@ protected function getSourceIterator()
 			? new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($this->source))
 			: new \ArrayIterator([new \SplFileInfo($this->source)]);
 
-		return new AcceptIterator($iterator, $this->acceptFiles);
+		return new \CallbackFilterIterator($iterator, function (\SplFileInfo $file) {
+			return substr($file->getBasename(), 0, 1) !== '.'  // . or .. or .gitignore
```
JanTvrdik: `$file->getBasename()[0]` should work fine assuming `getBasename()` cannot be empty

milo:
Updated. Imho, ready to merge. Some other objectives or updates?

milo:
GitHub closed the PR when I accidentally pushed `master` into `pull-php54` branch. Pushing commits back didn't help and PR now looks like 0 commits.

Merged:
- 077c3a940c26e251517a9f94d20b6d5a5985c39f
- db9688f40b666276579903e19ac852a2373354ac
- d7aad9ea7bcb9f06605d4c0b4d3a7c05b336001d
- 33fb94a711c4bba0a77287ca9904b04585bf48c4
- f836dbbe2d304d9fb687ab5a9148589cc5c41791
- c1bdc6fc106efb820fa901e19598d28f6a45efd9
- 8bd9db7f1c4a2e3acabeabcdf94c361009134fe8

PR #262, opened 2015-10-17
CliTester: explicit html_errors=off

milo:
It still can be overriden by `-d`.

The `html_output` is not useful in console output too I guess. But it would be always displayed in console like `PHP 5.6.13-0+deb8u1 | php-cgi -d html_errors=off | 8 threads`.


milo:
Merged 3d3060fde8e525ed5f434f35b96fd8d6c71fa23b

PR #270, opened 2015-11-02
TestCase: error handler is registered only once

milo:
See the commit message.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -18,6 +18,13 @@ class TestCase
 		METHOD_PATTERN = '#^test[A-Z0-9_]#';
 
 
+	/** @var bool */
+	private $handleErrors = FALSE;
+
+	/** @var callable|NULL|FALSE */
+	private $prevErrorHandler = FALSE;
+
+
 	/**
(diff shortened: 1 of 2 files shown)


src/Framework/TestCase.php
```diff
@@ -105,26 +112,33 @@ public function runTest($method, array $args = NULL)
...
+
+			$prev = set_error_handler(function ($severity) use ($me, & $prev, & $handleErrors) {
+				if ($handleErrors === TRUE && ($severity & error_reporting()) === $severity) {
+					$handleErrors = FALSE;
+					$rm = new \ReflectionMethod($me, 'tearDown');
+					$rm->setAccessible(TRUE);
+
+					set_error_handler(function() {});  // mute all errors
+					$rm->invoke($me);
+					restore_error_handler();
```
milo: This is the similar situation. User can setup own error handler in `tearDown`. But the `$handleErrors` prevents recursion now.

Mikulas:
:+1:, nextras/dbal does not fail anymore https://travis-ci.org/nextras/dbal/builds/93321753

PR #277, opened 2015-11-27
Dumper: object hash

tsusanka:
If `Assert:same()` is used to compare two similiar-looking objects and the assert fails, the Tester generates two outputs which shows no difference at all.

This PR prints object's hash to distinguish the objects.

### DIFF
--- a/src/Framework/Dumper.php
+++ b/src/Framework/Dumper.php
@@ -102,7 +102,7 @@ class Dumper
 			$line .= '(' . $object->format('Y-m-d H:i:s O') . ')';
 		}
 
-		return $line . '(#' . substr(md5(spl_object_hash($object)), 0, 4) . ')';
+		return $line . '(' . self::hash($object) . ')';
 	}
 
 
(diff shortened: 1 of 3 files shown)


Mikulas:
:+1: good idea, I've definitely have seen identical output before and got stuck on it for a while.

Why do you hash the object hash though?

dg:
What about non-stdClass objects?

tsusanka:
@Mikulas I got inspired by https://github.com/nette/tester/blob/master/src/Framework/Dumper.php#L105.  The `sql_object_hash()` returns "similiar" content to different objects, so it can't be easily stripped. See [function's comment](https://secure.php.net/manual/en/function.spl-object-hash.php#112199).

@dg this approach should work with non-stdClass objects as well, am I mistaken?

```bash
$ php
> class A {}
> $a = new A();
> echo spl_object_hash($a);
000000003ae624ea000000003019d5db
```

Majkl578:
> this approach should work with non-stdClass objects as well

It's in `$class === 'stdClass'`branch.

JanTvrdik:
@tsusanka awesome! Maybe return "$class::__set_state(/* #$objHash*/ array($out))" for other classes as well.

tsusanka:
Oh yeah, of course! Fixed.

milo:
@tsusanka :+1:  Could you please:
- separate it e.g. into `private static function hash()` (the `#` included in return value)
- squash it into single commit with message like `Dumper: print object hash into output dumps`

tsusanka:
Thx for the instructions. Done.

milo:
@tsusanka Thank you!

PR #279, opened 2015-12-29
Support for PHPDBG

milo:
This is my idea how to add support for PHPDBG interpreter for now.

The whole interpreter's interface would deserve refactoring, but it will be with a big BC break.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ matrix:
     allow_failures:
         - php: hhvm
 
+    include:
+        - php: 7.0
+          env: TESTER_PHP_BIN=phpdbg
+
     exclude:
         - php: 5.3
           env: TESTER_PHP_BIN=php-cgi
--- a/src/Runner/ZendPhpInterpreter.php
+++ b/src/Runner/ZendPhpInterpreter.php
@@ -7,6 +7,8 @@
 
 namespace Tester\Runner;
 
+use Tester\Helpers;
+
 
 /**
  * Zend PHP command-line executable.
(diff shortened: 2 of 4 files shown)


milo:
Ref krakjoe/phpdbg#149

Mikulas:
This does not support phpdbg code coverage though, does it? That was kind of the whole point of the original patch, mostly because phpdbg cov is at least twice as fast as xdebug's.

JanTvrdik:
This is useless by itself.

milo:
This PR is reaction to my [comment](https://github.com/nette/tester/pull/258#issuecomment-145285584), it is not your PR replacement.

Majkl578:
This looks like a dirty hack, phpdbg SAPI is completely different from normal CLI (e.g. command line arguments).
Coverage was the only reason why phpdbg was actually needed, because it is much faster. (Btw on my project, xdebug needs 90 seconds to complete, phpdbg needs only 14 seconds, both with PHP 7.0.2 and PHPUnit 5.)

milo:
@Majkl578 

> Coverage was the only reason why phpdbg was actually needed

Yes. And I agree that CC is faster. So what's the problem?

I didn't like the `ZendPhpDbgInterpreter` by @JanTvrdik. Imho, too much code is duplicated. Seems to me that this is more proper way for now. And as I wrote in this PR text, Interpreters would deserve refactoring.

My idea was: merge this and focus only on CC in #258.

milo:
Idea changed :)

Issue #278, opened 2015-12-17
Enforce .phpt file extension?


tsusanka:
Assume having two tests in one directory.

One `FirstTest.phpt`: 

``` php
// ...
class FirstTest extends Tester\TestCase
{
    public function testOk()
    {
        Assert::true(TRUE);
    }
}
```

and `AnotherTest.php`:

``` php
class AnotherTest extends Tester\TestCase
{
    public function testNotOk()
    {
        Assert::true(FALSE);
    }
}
```

You run them with all the other tests using `tester tests/` and you're super happy, since everything passes.

But wait, that's strange. The `AnotherTest` should fail! So you run `tester tests/AnotherTest.php` and it fails. What?! Now you're super-confused. Finally you realize `AnotherTest.php` is not run in the bulk, because it has accidentally wrong extension - `php` instead of `phpt`.

---

How about enforcing the `phpt` extension when running a single test? If I run `tester tests/AnotherText.php` and I get "no tests to run" I know something is wrong.

I'm not sure this is a good solution though, anybody has a better one?

fprochazka:
@tsusanka this could be added to the `Environment::setup()` method, good idea :+1:

Mikulas:
Plus one for addressing this issue, but overall
:-1: create an acceptance test that checks test files are named properly.

Majkl578:
Or rather maybe stop strictly enforcing `.phpt`?

tsusanka:
> Or rather maybe stop strictly enforcing .phpt?

You mean running all php/phpt files found in the directory? Bear in mind bootstraps, scripts etc.

Majkl578:
E.g. PHPUnit looks for files matching `*Test.php` or `*.phpt` by default. It does not force you to use specific format/extension.

tsusanka:
I like that. Some might have reasons not to use `.phpt`, for example [www.reviewable.io](reviewable.io) does not color highlight `.phpt` files as it does `.php` (which is a bug on their side, but still).

dg:
But this does not solve the problem.

tsusanka:
It does not, but it significantly helps. We can as well do both. Change the behavior as @Majkl578 suggests and enforce it during single tests as well.

dg:
The whole nette is not a single test, whose name ends with the `test`. How this helps me when I use accidentally wrong extension?

tsusanka:
[...]

PR #282, opened 2015-12-30
"-n" option ignores custom configuration

analogic:
Assume that getCommandLine() returns "/usr/bin/php -c /etc/php5/cli/php.ini". Adding " -n" option to end of string silently ignores "-c" setting (I was unable to load json extension originaly). I dont know if i am missing something but its working for me, also hhvm has no -n option apparently...

### DIFF
--- a/src/Runner/Job.php
+++ b/src/Runner/Job.php
@@ -83,7 +83,7 @@ class Job
 		putenv(Environment::COLORS . '=' . (int) Environment::$useColors);
 		$this->proc = proc_open(
 			$this->interpreter->getCommandLine()
-			. ' -n -d register_argc_argv=on ' . \Tester\Helpers::escapeArg($this->file) . ' ' . implode(' ', $this->args),
+			. ' -d register_argc_argv=on ' . \Tester\Helpers::escapeArg($this->file) . ' ' . implode(' ', $this->args),
 			array(
 				array('pipe', 'r'),
 				array('pipe', 'w'),


milo:
In which PHP version? The `-n` was there always and always worked for me since 5.3 to 5.6.

The purpose is to keep testing environmet clean as possible.

analogic:
PHP 5.5.9-1ubuntu4.14 (cli) - its standard ubuntu package with default config

Understood but I have not found any other way to run test with json_encode/decode. JSON extension is simply not loading

analogic:
Well php -c  /etc/php5/cli/php.ini -c /etc/php5/cli/conf.d -n works ok, but multiple "-c" options are not working for tester. Its seems to me that hardcoded "-n" is awful idea

milo:
This is a PHP issue a little bit. Problem is, that PHP is compilled to load `/etc/php/cli/conf.d/*.ini`. The `-n` disables this behaviour and there is no way how to restore it by command line option.

Take a look at this [doc chapter](https://tester.nette.org/en/#toc-own-php-ini). The `-n` is needed in tester.

analogic:
So please help me, how to use standard system configured php without problems? It seems deliberately not possible. 

Is clean php cleaness really issue? Ive used phpunit for years without any problems, maintaining another php.ini because json extension only adds another awkwardness bit to Nette/Tester for me...

analogic:
So is there possibility to change "-c" attribute to REPEATABLE at least?

milo:
> Is clean php cleaness really issue?

Yes. One issue is a clean environment. Another is possibility to run multiple versions of PHP.

> adds another awkwardness

Which are the previous? Maybe we can help you.

> So is there possibility to change "-c" attribute to REPEATABLE at least?

Maybe yes. Have to think about.

> So please help me, how to use standard system configured php without problems?

Not so easy due to PHP behaviour. If it is only about json, you can:
```
tester -c /etc/php/cli/php.ini -d extension=json.so tests
```

analogic:
Thanks for "-d", it should work but still it would be nice to be able opt-out to system configured php. Should i close this?

#278 was another "awkwardness" for me too but overall nette/tester seems nicer than phpunit

PR #283, opened 2016-01-01
PHP interpreters refactoring

milo:
This is my idea about PHP interpreters refactoring.

It is a lot of code and actually, all could be done in only one universal interpreter. But this seems more flexible to me.

The commit with the `info.php` brings new possibilities and reflects PHP environment more precisely.

I'm planning to use the `hasExtension()` method for `@phpExtension` annotation. It will speeds-up tests skipping and it will be shorter to write.


jkuchar:
@milo How stable is this, I would like to implement #294, should I base implementation on top of this or on top of master?

milo:
@jkuchar Unstable. I'll refactor it masivelly.

milo:
I updated the PR title and description. I would like to ask for review.
//cc @dg

milo:
Oh. Now I realized one important thing. That version detection by running `info.php` is too late. The command line arguments may not be compatible with current version. Hmmm...

src/Runner/Interpreters/AbstractInterpreter.php
```diff
@@ -0,0 +1,137 @@
...
+		return $this->info->version;
+	}
+
+
+	/**
+	 * @return bool
+	 */
+	public function canCodeCoverage()
+	{
+		return in_array('xdebug', $this->info->extensions,TRUE);
```
Majkl578: This should not be here, e.g. HHVM has XDebug but does not support coverage in a compatible way (flags do not exist).

src/Runner/Interpreters/ZendPhpCliInterpreter.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Zend PHP CLI executable.
+ */
+class ZendPhpCliInterpreter extends AbstractInterpreter
+{
+
+	public function __construct($path, $args = NULL)
+	{
+		parent::__construct($path, ' -n' . $args);
```
Majkl578: If args are specified, space will be missing after `-n` (applies to all variants above/below).

src/Runner/TestHandler.php
```diff
@@ -113,7 +113,7 @@ private function initiatePhpVersion($version, PhpInterpreter $interpreter)
 
 	private function initiatePhpIni($value, PhpInterpreter $interpreter)
 	{
-		$interpreter->arguments .= ' -d ' . Helpers::escapeArg($value);
+		$interpreter->addArguments(' -d ' . Helpers::escapeArg($value));
```
Majkl578: The space at the beginning looks a bit weird, this should be handled internally by interpreter itself.

src/Runner/PhpInterpreter.php
```diff
@@ -24,7 +29,7 @@ function getVersion();
 	/**
 	 * @return bool
 	 */
-	function hasXdebug();
+	function canCodeCoverage();
```
dg: canMeasureCodeCoverage, canMeasureCoverage, canAnalyseCoverage?

hrach: canDoCodeCoverage?

JanTvrdik: :+1: for `canMeasureCodeCoverage`.

---

Wouldn't it be better to ask the `CodeCoverage\Collector` whether he can do it?

src/Runner/CliTester.php
```diff
@@ -176,18 +177,17 @@ private function createPhpInterpreter()
...
-			$this->interpreter = new HhvmPhpInterpreter($this->options['-p'], $args);
+			$this->interpreter = new Interpreters\HhvmPhpInterpreter($this->options['-p'], $args);
 		} elseif (strpos($output, 'phpdbg') !== FALSE) {
-			$this->interpreter = new ZendPhpDbgInterpreter($this->options['-p'], $args);
+			$this->interpreter = new Interpreters\ZendPhpDbgInterpreter($this->options['-p'], $args);
+		} elseif (strpos($output, 'cgi') !== FALSE) {
+			$this->interpreter = new Interpreters\ZendPhpCgiInterpreter($this->options['-p'], $args);
 		} else {
-			$this->interpreter = new ZendPhpInterpreter($this->options['-p'], $args);
+			$this->interpreter = new Interpreters\ZendPhpCliInterpreter($this->options['-p'], $args);
```
JanTvrdik: move to elseif with strpos=cli + add else with exception

src/Runner/Interpreters/AbstractInterpreter.php
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+
+	/**
+	 * @return bool
+	 */
+	public function isCgi()
+	{
+		return FALSE;
+	}
```
JanTvrdik: This looks like a method we should try to get rid off.

src/Runner/Interpreters/HhvmPhpInterpreter.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * HHVM command-line executable.
+ */
+class HhvmPhpInterpreter extends AbstractInterpreter
+{
+
+	public function __construct($path, $args = NULL)
+	{
+		parent::__construct($path, ' --php -n -d hhvm.log.always_log_unhandled_exceptions=false' . $args); // HHVM issue #3019
```
JanTvrdik:
~~~php
$args = " --php -n -d hhvm.log.always_log_unhandled_exceptions=false $args" // HHVM issue #3019
parent::__construct($path, $args);
~~~

src/Runner/TestHandler.php
JanTvrdik: Have you considered moving the logic for arguments handling to interpreters, i.e. sth. like

~~~php
$interpreter->setPhpIniPath($value);
~~~

src/Runner/PhpInterpreter.php
```diff
@@ -12,6 +12,11 @@
 {
 
 	/**
+	 * @param  string
+	 */
+	function addArguments($args);
```
JanTvrdik: I think that it would better to remove this from the interface and make it `protected` on `AbstractInterpteter`.

JanTvrdik:
@milo Overall very good job. I would try get rid off `addArguments` and `isCgi` methods.

src/Runner/PhpInterpreter.php
Majkl578: supportsCodeCoverage?

dg: bikeshedding :-)

src/Runner/Interpreters/AbstractInterpreter.php
milo: The HHVM interpreter overloads that. Actually, it could be an abstract method.

src/Runner/Interpreters/ZendPhpCliInterpreter.php
milo: I tried to normalized it across the Tester. The space was always [required](https://github.com/nette/tester/blob/5c4f5ec69d6894cd92bbeb1a216a63914f798761/src/Runner/ZendPhpInterpreter.php#L58). Now it is consistent.

Imho, it is not so important. Interpreters are "internal" classes.

src/Runner/TestHandler.php
milo: > The space at the beginning looks a bit weird

It looks. As I wrote sooner, it is an internal code. I was thinking about passing arguments as an unescaped array like `addArguments(['-d', 'some value'])`.

> Have you considered moving the logic for arguments handling to interpreters

I did. But it needs `-d something` (passed by user) parsing which I would like to prevent if not needed. And the name would be probably `addPhpIniOption()` since one option can be set multiple times.

src/Runner/PhpInterpreter.php
milo: Hehe, I was expecting this discussion :o) The `canCodeCoverage` is nonsence and I didn't get better idea. The `canMeasureCodeCoverage` seems good.

src/Runner/Interpreters/AbstractInterpreter.php
milo: Some idea how? It is needed to know, if stdout contains HTTP code with headers or not.

src/Runner/PhpInterpreter.php
milo: I added this method to the interface because of former `$arguments` property. The `addPhpIniValue()` would solve it. Maybe in `addPhpIniValue('name=value')` format.

milo:
@JanTvrdik Thanks. One step how to remove `isCgi()` is moving Job's run logic into interpreter.

src/Runner/TestHandler.php
JanTvrdik: > But it needs -d something (passed by user) parsing

You can just split it on the first `=` sign, or not?

milo: Hmm. Now I cannot find it. I recall there were more options for format. Now manual talks about `=` only. Maybe they dopped it. Anyway, `addPhpIniOption('name=value')` could be fine too.

src/Runner/Interpreters/AbstractInterpreter.php
JanTvrdik: It depends on how far are you willing to go...

You can rename the method based on what it actually means / is used for, i.e. sth like `isHttpMode`
or integrate what is the `isCgi()` is being used for, i.e. add sth like `parseRawOutput`
or Integrate most of Job's logic

milo:
Refactored. I'm not sure about bdaf6292c9ea747ea50c794c9c5384812a74c8be. Code looks more messy, but it encapsulates `Helpers::escapeArg()` usage inside the interpreters.

milo:
When no more comments on it, I'll merge it as is.

dg:
@milo please wait a moment

milo:
@dg Sure. I can pospone it if you wish. I just wanted to merge it before next Tester changes, because it is quite big change and rebasing is a pain.

dg:
In HHVM -d and -c are supported? The only difference is `-n` versus `--no-config`?

dg:
1) PhpInterpreter is (nearly) dump data object with no logic, I think that there is no need to have 1 interface, 1 abstract class and 3 implementations for data object.
2) In CliTester::createPhpInterpreter() is first part of PhpInterpreter factory, in each implementation is second part and it should be moved together

What do you think about #306 ? +183 −401 versus +378 −371 :-)

milo:
@dg :+1:

milo:
Closing in favour of #306

PR #288, opened 2016-02-12
CodeCoverage: HTML template facelift

milo:
Reworking #237, colors by GitHub.

Colors are not 1:1 with GitHub, it would need better granularity than `highlight_string()`, tokenize. Anyway, I'm not sure we want the look same as GitHub. When I had opened more windows, I'm loosing focus what is a code coverage and what is the GitHub.

But it is not about colors only. After this refactoring, output HTML is 752kB instead of 1.23MB, so maybe worth it.

Before some color fight, post some CSS propose please, e.g. on fiddle, if possible.

Example is here: https://hoola.cz/files/coverage.288-1.html

### DIFF
--- a/src/CodeCoverage/Generators/template.phtml
+++ b/src/CodeCoverage/Generators/template.phtml
@@ -9,45 +9,72 @@
 
 	<style type="text/css">
 	html {
-		font: 16px/1.5 sans-serif;
-		border-top: 4.7em solid #F4EBDB;
+		font: 14px/1.5 Verdana,"Geneva CE",lucida,sans-serif;
+		border-top: 4.7em solid #f4ebdb;
 	}
 
 	body {
 		max-width: 990px;
 		margin: -4.7em auto 0;
-		background: white;
+		background: #fcfaf5;
 		color: #333;
 	}
 
+	footer {
+		margin-left: .5em;
+	}
+
 	h1 {
+		font-family: "Trebuchet MS","Geneva CE",lucida,sans-serif;
 		font-size: 1.9em;
-		margin: .5em 0 1.5em;
-		color: #7A7772;
+		margin: .5em .5em 1.5em;
+		color: #7a7772;
 		text-shadow: 1px 1px 0 white;
 	}
 
 	div.code {
-		font: 14px/1.2 Consolas, monospace;
-		background: #FDF5CE;
-		padding: .4em .7em;
+		background: white;
 		border: 1px dotted silver;
-		overflow-x: auto;
 		display: none;
+		color: #333;
+		position: relative;
+		overflow: auto;
 	}
 
-	span.line {
-		color: #9F9C7F;
-		font-weight: normal;
-		font-style: normal;
+	code,
+	div.code {
+		font: 13px/1.3 monospace;
+		padding: .4em .7em;
 	}
 
-	.t {
-		background: #99f999;
+	div.code div {
+		white-space: pre;
 	}
 
-	.u {
-		background: #f9ac9e;
+	div.code a {
+		color: #c0c0c0;
+	}
+
+	div.code a:hover {
+		color: inherit;
+		font-weight: bold;
+		text-decoration: none;
+	}
+
+	code {
+		white-space: nowrap;
+		position: absolute;
+		top: 0;
+	}
+
+	a {
+		color: #006aeb;
+		text-decoration: none;
+	}
+
+	a:active,
+	a:hover {
+		text-decoration: underline;
 	}
 
 	td {
@@ -64,14 +91,14 @@
 	}
 
 	.bar {
-		border: 1px solid #ACACAC;
+		border: 1px solid #acacac;
 		background: #e50400;
 		width: 35px;
 		height: 1em;
 	}
 
 	.bar div {
-		background: #1A7E1E;
+		background: #1a7e1e;
 		height: 1em;
 	}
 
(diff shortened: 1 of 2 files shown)


Mikulas:
Definitely an improvement over the current style, so for that +1.

Are overflows intentionally overflowing or is it an oversight? It's probably better to show the code rather than cropping it to the box, but it does not look that great. Forced scrolling might be a viable option (which is what github [does](https://github.com/nette/tester/blob/master/src/CodeCoverage/Generators/template.phtml)).

screenshot 2016-02-12 23 18 55
(example of `CodeCoverage\Generators\template.phtml`)

For the record, I don't really care about output size, not of the built html anyway.

Proposal: clicking a file should actually change fragment in url to simplify sharing between coworkers. Similarly so should line numbers. This could be sent in a future PR though.

dg:
Isn't it better to use color scheme similar to Tester API doc? https://api.nette.org/tester/1.7.0/source-Framework.Assert.php.html

milo:
@Mikulas Overflow was intentional.

@dg Maybe it would look better, but when trying it, there is almost everything `default` or `keyword` by `highlight_string()`. And it looses clarity.

milo:
I'm thinking about tokenizing, but have to measure how slowet it would be.

milo:
It tried own highlighter. It is about 31% slower, mainly due to tokenize.
- with [GitHub theme](https://hoola.cz/files/coverage.288-2.html)
- with [Nette ApiGen colors](https://hoola.cz/files/coverage.288-3.html)

dg:
Wow, own highlighter!

dg:
I am perfecty fine with GitHub theme and with Nette theme too. For consistency and in order to not look like GitHub rip, I am more for Nette colors. And links (filenames) should match palette too.

I see major importance of highlighter in differentiate between tested lines / untested lines / class & method declaration / comments. Everything other like highlight of variables etc is actually rather disturbing. So default `highlight_string()` with only 'default' and 'keyword' is probably for this case adequate.

milo:
> Wow, own highlighter!

But still, it needs more granularity! ;) It is just test digged out from one of my old lib. I'm planning to revert it.

> Everything other like highlight of variables etc is actually rather disturbing

Have to admit. I'm using it whole the day and it is really distracting.

milo:
What about [this approach](https://hoola.cz/files/coverage.288-4.html)?

dg:
Nice!

dg:
Overflow seems like a bug, especially now when code has bigger font, why is it intentional?

milo:
Well, I dont like vertical scrolling to much. But now it looks better with scroll bars 
- https://hoola.cz/files/coverage.288-5.html

milo:
Rebased and squashed. Imho ready to merge.
- [before](https://hoola.cz/files/coverage.html) 1243kB
- [now](https://hoola.cz/files/coverage.288-6.html) 891kB

src/CodeCoverage/Generators/template.phtml
```diff
@@ -98,49 +134,39 @@
 		<tr<?= $info->class ? " class='$info->class'" : '' ?>>
 			<td class="number"><small><?= $info->coverage ?> %</small></td>
 			<td><div class="bar"><div style="width: <?= $info->coverage ?>%"></div></div></td>
-			<td><a href="#fragment<?= $id ?>" onclick="toggle(this); return false"><span><?= $info->name ?></span></a></td>
+			<td><a href="#F<?= $id ?>" onclick="toggle(this.href); return false"><span><?= $info->name ?></span></a></td>
```
dg: `<span>` around `$info->name` seems useless

src/CodeCoverage/Generators/template.phtml
```diff
@@ -9,45 +9,72 @@
...
-		overflow-x: auto;
 		display: none;
+		color: #333;
+		position: relative;
+		overflow: auto;
+	}
+
+	code,
+	div.code {
+		font-family: monospace;
```
dg: IMHO something like `font: 13px/1.3 monospace;` should make it better.

dg:
There is problem with short hightlighted lines, when you scroll code right. The solution is:

```html
<div class="code">  (remove here position:relative, it will become useles)
NEW   <div class="content"> // float: left; position: relative
NEW      <div class="lines">  // min-width: 100%; position: absolute;
                    <div><a href='#F1L1' id='F1L1'> 1</a>…  //original divs with lines
              </div>
              <code> ..... </code>   // position: relative (instead of absolute), left padding 3em (instead of left: 3em), maybe display: block
          </div>
</div>
```

milo:
@dg First things fixed. But I'am not able to handle the short lines issue. There is always some ugly margin or white-space issue. Frontend help is very appreciated :) Or I'll elaborate later.

src/CodeCoverage/Generators/template.phtml
```diff
@@ -176,18 +176,21 @@
...
+	document.body.addEventListener('click', function (e) {
+		if (e.target.className === 'toggle') {
+			var el = document.getElementById(e.target.href.split('#')[1]);
+			if (el.style.display === 'block') {
+				el.style.display = 'none';
+			} else {
+				el.style.display = 'block';
+			}
+			e.preventDefault();
 			return false;
```
dg: return false can be removed

src/CodeCoverage/Generators/template.phtml
```diff
@@ -176,18 +176,21 @@
...
-		} else {
-			el.style.display = 'block';
 		}
-	}
+	});
 
 	if (window.location.hash.length) {
-		toggle(window.location.hash.split('L')[0]);
+		document.getElementById(window.location.hash.replace(/^#|L\d+$/g, '')).style.display = 'block';
 	}
```
dg: more error prone `if (el = document.getElementById(window.location.hash.replace(/^#|L\d+$/g, ''))) { el.style.display = 'block'; }`

dg:
Ok, merge it and I'll try to solve line issue :-)

milo:
> I'll try to solve line issue

Thank you! Next beer is on me :)

dg:
:beers:

PR #289, opened 2016-02-20
Assert: added noException (opposite to exception/throws)

greeny:
/cc @milo 

Feature request, no BC breaks.

Handy method to simplify pieces of code like:

```php
try {
    someCode();
} catch (Exception $e) {
    Assert::fail();
}
```

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -316,6 +316,26 @@ class Assert
 	{
 		return self::exception($function, $class, $message, $code);
 	}
+	
+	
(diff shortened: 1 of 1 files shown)


fprochazka:
You don't really have to do that, the test will just fail if an exception is thrown. Also, the fail message should contain info about what exception was thrown.

milo:
There is an Assert:noError() which catches exceptions too.

dg:
This duplicates `noError()`

Issue #291, opened 2016-02-24
FileMock does not handle 'w' mode


danielkurecka:
Consider this code:
```php
$filename = Tester\FileMock::create('foo');
fopen($filename, 'w');

// content of the file should be empty now, but it's not
var_dump(file_get_contents($filename));
```

milo:
Forget to mention that fixed by #292

PR #292, opened 2016-02-24
FileMock fixes

milo:


src/Framework/FileMock.php
```diff
@@ -40,16 +48,51 @@ public static function create($content, $extension = NULL)
...
+		if (!self::$registered) {
+			stream_wrapper_register(self::PROTOCOL, __CLASS__);
+			self::$registered = TRUE;
+		}
+	}
+
+
 	public function stream_open($path, $mode)
 	{
+		if (!preg_match('#^([rwaxc]).*?(\+)?#', $mode, $m)) {
```
milo: Not sure, how PCRE should be strict. I tried, that first letter must match mode, others can be mess.

milo:
@dg What do you think about this approach? It is mainly fix for #291. Not sure it shoud be 1.7.x included.

src/Framework/FileMock.php
```diff
@@ -40,16 +48,51 @@ public static function create($content, $extension = NULL)
 	}
 
 
+	public static function register()
```
milo: Purpose: when someone want to use mock:// before `FileMock::create()`, e.g. testing with `mock://not-exist-yet`.

milo:
Not sure why it fails in here (https://github.com/nette/tester/blob/84bcf2514baa8e4775cb27461de7bac3dcfae519/tests/CodeCoverage/CloverXMLGenerator.phpt#L27)

src/Framework/FileMock.php
```diff
@@ -40,16 +48,51 @@ public static function create($content, $extension = NULL)
...
+	{
+		if (!self::$registered) {
```
dg: Maybe can be replaced with stream_get_wrappers() https://github.com/nette/safe-stream/commit/4b9bb3266a537e59b10a2932f05dca47333420fc

dg:
Great :+1:

PR #293, opened 2016-03-03
TestCase: possibility of named sets in dataprovider

stekycz:
Minor features which brings easier distinguishing which data the test is failing for. Useful for larger and more complicated datasets at the input.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -98,8 +98,13 @@ class TestCase
 				if (!is_array($res) && !$res instanceof \Traversable) {
 					throw new TestCaseException("Data provider $provider() doesn't return array or Traversable.");
 				}
-				foreach ($res as $set) {
-					$data[] = is_string(key($set)) ? array_merge($defaultParams, $set) : $set;
+				foreach ($res as $setName => $set) {
+					$set = is_string(key($set)) ? array_merge($defaultParams, $set) : $set;
+					if (is_string($setName)) {
+						$data[$setName] = $set;
+					} else {
+						$data[] = $set;
+					}
 				}
 			}
 
@@ -126,7 +131,7 @@ class TestCase
 		}
 
 
-		foreach ($data as $params) {
+		foreach ($data as $setName => $params) {
 			try {
 				$this->setUp();
 
@@ -152,7 +157,7 @@ class TestCase
 				$this->tearDown();
 
 			} catch (AssertException $e) {
-				throw $e->setMessage("$e->origMessage in {$method->getName()}(" . (substr(Dumper::toLine($params), 1, -1)) . ')');
+				throw $e->setMessage("$e->origMessage in " . (is_string($setName) ? "\"" . $setName . "\": " : "") . "{$method->getName()}(" . (substr(Dumper::toLine($params), 1, -1)) . ')');
 			}
 		}
 	}
--- a/tests/Framework/TestCase.dataProvider.phpt
+++ b/tests/Framework/TestCase.dataProvider.phpt
@@ -18,6 +18,14 @@ class MyTest extends Tester\TestCase
 		];
 	}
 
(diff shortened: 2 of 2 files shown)


src/Framework/TestCase.php
```diff
@@ -97,8 +97,13 @@ public function runTest($method, array $args = NULL)
...
 				}
-				foreach ($res as $set) {
-					$data[] = is_string(key($set)) ? array_merge($defaultParams, $set) : $set;
+				foreach ($res as $setName => $set) {
+					$set = is_string(key($set)) ? array_merge($defaultParams, $set) : $set;
+					if (is_string($setName)) {
+						$data[$setName] = $set;
+					} else {
+						$data[] = $set;
+					}
```
milo: @stekycz All original keys can be preserved in `$data[]`, not only strings.

src/Framework/TestCase.php
```diff
@@ -167,7 +172,7 @@ public function runTest($method, array $args = NULL)
 				}
 
 			} catch (AssertException $e) {
-				throw $e->setMessage("$e->origMessage in {$method->getName()}" . (substr(Dumper::toLine($params), 5)));
+				throw $e->setMessage("$e->origMessage in " . (is_string($setName) ? "\"" . $setName . "\": " : "") . "{$method->getName()}" . (substr(Dumper::toLine($params), 5)));
 			}
```
milo: I'm thining about better message composition.

tests/Framework/TestCase.dataProvider.phpt
```diff
@@ -59,6 +73,12 @@ class MyTest extends Tester\TestCase
 	{
 		Assert::true(FALSE);
 	}
+
+	/** @dataProvider dataProviderNamedSets */
+	public function testAssertionNamedSets()
+	{
+		Assert::true(false);
+	}
```
milo: `FALSE`

tests/Framework/TestCase.dataProvider.phpt
```diff
@@ -84,6 +104,14 @@ Assert::same(array(
 
 
 $test = new MyTest;
+$test->run('testDataProviderNamedSets');
+Assert::same(array(
+	'MyTest::dataProviderNamedSets',
```
milo: You can use short arrays notation now :)

src/Framework/TestCase.php
stekycz: @milo Unfortunately I can't because it breaks possibility of multiple data providers. The string key must be unique because it is explicit value however two different data providers without naming returns array with indexes `0, 1, 2, ...` so it results in wrong data set. You can try it on your own ;-)

stekycz: @milo Unfortunately I can't because it breaks possibility of multiple data providers. The string key must be unique because it is explicit value however two different data providers without naming returns array with indexes `0, 1, 2, ...` so it results in wrong data set. You can try it on your own ;-)

tests/Framework/TestCase.dataProvider.phpt
stekycz: Updated

tests/Framework/TestCase.dataProvider.phpt
stekycz: Fixed

src/Framework/TestCase.php
stekycz: Do you mean wording, format or code?

src/Framework/TestCase.php
milo: Actually, the same string may replace previous dataset too.

stekycz: Yes it can. However the probability of that is lower then for regular array. The same string name must be specified explicitly however same array index is quite hard to find where the problem is.

stekycz:
Rebased

PR #295, opened 2016-03-16
added .gitattributes

JanTvrdik:


.gitattributes
```diff
@@ -0,0 +1,5 @@
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+tests/ export-ignore
+*.sh eol=lf
```
dg: Why *.sh ?

fprochazka: `.sh` is relevant to linux only, and nobody on linux is using windows line endings?

dg: There is no *.sh file

PR #298, opened 2016-03-20
Allow manually save partial code coverage to free memory

JanTvrdik:
phpdbg code coverage is very fast but it also consumes a lot of memory because it stores information about every opcode (PHP low-level instruction) executed. This can be alleviated by allowing programmer to manually save partial code coverage to file.


JanTvrdik:
BTW: Can anyone think of a reason why the file is open in `a+` mode instead of `c+`?

dg:
Because c+ did not exist in PHP 5.2.0

JanTvrdik:
https://github.com/nette/tester/pull/299 is required for new tests for this

JanTvrdik:
rebased

JanTvrdik:
ping

milo:
Thank you.

PR #299, opened 2016-03-20
FileMock: incorrect handling of position

JanTvrdik:
When writing tests for https://github.com/nette/tester/pull/298 I discovered bug in FileMock implementation which took me several hours to isolate and I still don't feel I fully understand it.

To sum up my observation (and I am still possible wrong as I was *too long too deep*): Position works differently in append mode – writing is always appended and position affect only reading and starts on zero (i.e. you need to track read and writing position independently). However even thought calling ftruncate should not move file pointer (see phpdoc), it does actually move the writing pointer in this case.

My guess is that you will need three numbers to handle all edge cases correctly instead of one.

### DIFF
--- a/src/Framework/FileMock.php
+++ b/src/Framework/FileMock.php
@@ -22,7 +22,13 @@ class FileMock
 	private $content;
 
 	/** @var int */
-	private $pos;
+	private $readingPos;
+
+	/** @var int */
+	private $writingPos;
+
+	/** @var bool */
+	private $appendMode;
 
 	/** @var bool */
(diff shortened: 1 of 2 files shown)


dg:
It is in documentation http://php.net/manual/en/function.fopen.php

> 'a+'	Open for reading and writing; … In this mode, fseek() only affects the reading position, writes are always appended.

JanTvrdik:
Hopefully fixed.

dg:
Thanks

PR #304, opened 2016-03-22
Another failure in FileMock 

JanTvrdik:


JanTvrdik:
Isolated to https://3v4l.org/gTXsR, seems like a PHP bug.

milo:
@JanTvrdik I thought it has been solved but it's still failing on 3v4al.org. Did you report PHP bug? I can't find any open or solved.

JanTvrdik:
@milo I don't remember opening any bug so it's possible that it has never been reported.

JanTvrdik:
Just found in PHP documentation

>  If you have opened the file in append (a or a+) mode, any data you write to the file will always be appended, regardless of the file position, and **the result of calling fseek() will be undefined.**

milo:
But in doc of `fopen()`

> `a` - Open for writing only; place the file pointer at the end of the file. If the file does not exist, attempt to create it. In this mode, **fseek() has no effect, writes are always appended**. 
> `a+` - Open for reading and writing; place the file pointer at the end of the file. If the file does not exist, attempt to create it. In this mode, **fseek() only affects the reading position, writes are always appended**.

The `a+` mode is about read/append. How would one can read.

I'll open PHP bug/doc issue.

milo:
Will see https://bugs.php.net/bug.php?id=76995

milo:
As I expected, it is a [problem of lower layer](https://bugs.php.net/bug.php?id=76995#1539187632). So, documentation issue.

Is a good idea to emit notice by `FileMock::stream_seek()` in `a+` mode? IMHO yes, it may notice programmer about this clumsy behaviour.

milo:
Still buggy in PHP 8.0

PR #306, opened 2016-04-17
PHP interpreters refactoring II.

dg:


dg:
Of course the question is whether distinct between HHVM, phpdbg and PHP by file name or by `--version`. IMHO file name sufficient, but it can be simply changed back. The point is that now it is done out of `CliTester`.

Majkl578:
> IMHO the file name is sufficient

It is not. HHVM is provided as /usr/bin/php on Debian-based systems through /etc/alternatives.

dg:
Ok, I'll change it

milo:
@dg Thanks for review and refactoring!

PR #308, opened 2016-04-26
TestCase: mute tearDown() exceptions when called after error in test

milo:


milo:
@dg May I ask for review?

dg:
I have absolutely no idea what it does, TestCase is out of my scope and especially this part is pure magic :-)

PR #309, opened 2016-05-03
BC breaks for v2.0

milo:
There are proposes for BC breaks :o) Every commit references own issue. I'll leave few comments among diffs.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -39,7 +39,7 @@ class TestCase
 			return $rm->getName();
 		}, (new \ReflectionObject($this))->getMethods())));
 
-		if (isset($_SERVER['argv']) && ($tmp = preg_filter('#(--method=)?([\w-]+)$#Ai', '$2', $_SERVER['argv']))) {
+		if (isset($_SERVER['argv']) && ($tmp = preg_filter('#--method=([\w-]+)$#Ai', '$1', $_SERVER['argv']))) {
 			$method = reset($tmp);
 			if ($method === self::LIST_METHODS) {
 				Environment::$checkAssertions = FALSE;
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -29,18 +29,17 @@ class TestCase
 	 * Runs the test case.
 	 * @return void
 	 */
-	public function run($method = NULL)
+	public function run()
 	{
-		$r = new \ReflectionObject($this);
-		$methods = array_values(preg_grep(self::METHOD_PATTERN, array_map(function (\ReflectionMethod $rm) {
-			return $rm->getName();
-		}, $r->getMethods())));
-
-		if (substr($method, 0, 2) === '--') { // back compatibility
-			$method = NULL;
+		if (func_num_args()) {
+			throw new \LogicException('Calling TestCase::run($method) is deprecated. Use TestCase::runTest($method) instead.');
 		}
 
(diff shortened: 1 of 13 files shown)


src/Framework/TestCase.php
```diff
@@ -29,35 +29,31 @@ class TestCase
...
-		}, $r->getMethods())));
-
-		if (substr($method, 0, 2) === '--') { // back compatibility
-			$method = NULL;
-		}
+		}, (new \ReflectionObject($this))->getMethods())));
 
-		if ($method === NULL && isset($_SERVER['argv']) && ($tmp = preg_filter('#(--method=)?([\w-]+)$#Ai', '$2', $_SERVER['argv']))) {
+		# TODO: Support 'php test.phpt testMethod' or 'php test.phpt --method=testMethod' only?
+		if (isset($_SERVER['argv']) && ($tmp = preg_filter('#(--method=)?([\w-]+)$#Ai', '$2', $_SERVER['argv']))) {
```
milo: Not sure about support manual args without `--method=`.

src/Framework/TestCase.php
```diff
@@ -70,6 +66,12 @@ public function run($method = NULL)
 	 */
 	public function runTest($method, array $args = NULL)
 	{
+		if (!method_exists($this, $method)) {
+			throw new TestCaseException("Method '$method' does not exist.");
```
milo: Actually this is not needed. The `ReflectionException` will be thrown. But this have a straightforward message.

src/Framework/TestCase.php
```diff
@@ -70,6 +66,12 @@ public function run($method = NULL)
...
+		} elseif (!preg_match(self::METHOD_PATTERN, $method)) {
+			throw new TestCaseException("Method '$method' is not a testing method.");
```
milo: This forces `runTest()` usage only for test methods. IMHO OK, any other public method can be called directly. But maybe, there can be some use case for it.

milo:
Output may look like:
```
$ src/tester tests
 _____ ___  ___ _____ ___  ___
|_   _/ __)( __/_   _/ __)| _ )
  |_| \___ /___) |_| \___ |_|_\  v2.0.x

Note: Default interpreter is CLI since Tester v2.0. It used to be CGI.
Note: No php.ini is used.
PHP 5.6.13 (cli) | php -n  | 8 threads

-- FAILED: tests\Framework\TestCase.annotationThrows.phpt
   Exited with error code 255 (expected 0)
   LogicException: Calling TestCase::run($method) is deprecated. Use TestCase::runTest($method).
```

dg:
:+1:

src/Framework/TestCase.php
```diff
@@ -29,35 +29,31 @@ class TestCase
...
+		}
+
 		$methods = array_values(preg_grep(self::METHOD_PATTERN, array_map(function (\ReflectionMethod $rm) {
 			return $rm->getName();
-		}, $r->getMethods())));
-
-		if (substr($method, 0, 2) === '--') { // back compatibility
-			$method = NULL;
-		}
+		}, (new \ReflectionObject($this))->getMethods())));
```
JanTvrdik: Extract to `private getTestMethods()`?

src/Framework/TestCase.php
dg: I am not using TestCase, but I think that it is not needed to have two ways how do the same think and it blocks first parameter.

src/Framework/TestCase.php
milo: IMHO not needed. It is used only once.

PR #310, opened 2016-05-04
The -o option can be used multiple times

milo:
For example:

```
vendor/bin/tester -o console -o tap:output.tap -o junit:output.xml -o log:output.log
```

Tester ASCII art logo is shown, only if output for stdout (without parameter) is not passed.


dg:
Would not it be easier to leave CommandLine as is, change parameter to repeatable and split it in loadOptions()?

milo:
@dg Probably yes. The `-o value:file/path` syntax seems ok? So, drop `CommandLine::ENUM` (this is the only one place) and parse it manually. Or add `CommandLine::NORMALIZER` like:
```php
CommandLine::NORMALIZER => function ($arg) {
    list($value, $path) = explode(':', $arg, 2);
    if (...) throw ...
};
```
I'll try it, how it will look like.

milo:
@dg What about such approach (last commit)?

dg:
I tried to find if there is any pattern how should these parameters look, but it seems that this is pole neorané … not standardized. Sometimes it is used this way:

```
vendor/bin/tester -o console -o tap -file output.tap -o junit -file output.xml -o log -file output.log
```

Which looks little bit weird. So syntax looks fine to me.

src/Runner/CliTester.php
```diff
@@ -96,21 +99,22 @@ private function loadOptions()
...
-    --coverage-src <path>        Path to source code.
-    -h | --help                  This help.
+    -p <path>                Specify PHP interpreter to run (default: php).
+    -c <path>                Look for php.ini file (or look in directory) <path>.
+    -d <key=value>...        Define INI entry 'key' with value 'val'.
+    -s                       Show information about skipped tests.
+    --stop-on-fail           Stop execution upon the first failure.
+    -j <num>                 Run <num> jobs in parallel (default: 8).
+    -o <console|tap|junit|log|none>
+                             Specify output format. Repeatable with parameter.
```
dg: `Specify one or more output formats with optional file name` is probably better that `Repeatable with parameter`.

dg:
In fact ENUM is used by Tester, try `tester --colors x` before and after.

milo:
> In fact ENUM is used by Tester, try tester --colors x before and after.

Ups, I was rash. I have to find a way how to preserve `console|tap|junit|log|none` at one place and do not collide with ENUM implementation. Or replace ENUM implementation by something similarly easy.

milo:
Closing in favour of #329

milo:
Closing in favour of #329

PR #311, opened 2016-05-07
Improve parameter names in type assert method

PunchyRascal:
so that it is clear what is the expected and what is the actual value

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -251,27 +251,27 @@ class Assert
 	 * Checks assertion.
 	 * @return void
 	 */
-	public static function type($type, $value, $description = NULL)
+	public static function type($expectedType, $actualValue, $description = NULL)
 	{
 		self::$counter++;
-		if (!is_object($type) && !is_string($type)) {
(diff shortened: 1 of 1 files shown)


Majkl578:
Could you also add phpDoc for parameters (which is missing now)? :)

dg:
:-1: in the same spirit we can change `count($count, $value)` to `count($expectedCount, $actionValue)` and it seems unnecessary.

PunchyRascal:
Unnecessary in what sense? I honestly had no idea what these parameters meant, until I looked into the code itself. I think that variable names like $value, $data, are in most contexts dangerously generic and folks who don't know the code can find it confusing. It is almost as using $string, $number, $a, $b, $c etc.

I added the phpDoc.

PunchyRascal:
I agree that other similarly affected methods should be refactored, too. I might look into it.

Majkl578:
> it seems unnecessar

As it's something that doesn't affect performance, but improves readability & user experience, I'm :+1: .

jkuchar:
I like it; I'm in.

PunchyRascal:
Maybe this PR will merge on its first birthday then? :-D

PR #314, opened 2016-05-15
Assert can find mismatch position when using patterns

JanTvrdik:
Killer feature when comparing large files, such as Tracy Bluescreen outputs.

It finds the longest pattern that can be matched, then the shortest string that matches such pattern with binary search and finally it finds the longest string that matches such pattern.


milo:
This is one of TODO in my head :) so :+1: Mine idea was adjust dumps to be easily comparable by some external diff. E.g. replace `%a%` pattern by spaces as long as possible.

JanTvrdik:
I originally thought that the implementation would be a LOT simpler as I imaged just two for loops (one for $pattern and one for $actual). But `Assert::match` is so slow, that it would takes minutes to find the mismatch offset.

JanTvrdik:
We may simplify the `*.actual` and `*.expected` files from `@outputMatch` as well but I'm not sure whether that's a good idea.

JanTvrdik:
@milo The last commit changes the behavior to possible match what you intended to do.

dg:
Wow

JanTvrdik:
Now with defense against :smiling_imp: patterns! :smile:

JanTvrdik:
@dg This is now ready to merge from my point of view.

dg:
This is awesome!

It gives even better results when right trim (ie. `$actual = preg_replace('#^[\t ]+(?=[\r\n])#m', '', $actual);`) is applied on strings that are passed to exception. Dumper::dumpException($e) show real point of first difference.

dg:
Btw: @milo is seems that stuff around match() is ready to be moved to a new class.

JanTvrdik:
> It gives even better results when right trim (…) is applied on strings

I changed `expandMatchingPatterns` to return [pattern, actual] with both values normalized.

milo:
:100: like it, I didn't forgot on this. But I would like to find time to get in deeper in it before merge.

JanTvrdik:
@milo Last time that happen [it took months](https://github.com/nette/tester/pull/258) to get it merge.

milo:
Yes, that's correct. Are you hurry on this?

JanTvrdik:
I like to have as little open PR as possible. As long as this is not merged it is very difficult for me to send another non-trivial PR to Tester. There are multiple interesting problems that could be solved as a next step but I can not do it before this gets merged.

src/Framework/Assert.php
```diff
@@ -509,6 +518,67 @@ public static function isMatching($pattern, $actual)
...
+	 * @return array
+	 * @internal
+	 */
+	public static function expandMatchingPatterns($pattern, $actual)
+	{
+		if (self::isPcre($pattern)) {
+			return [$pattern, $actual];
+		}
+
+		$parts = preg_split('#(%)#', $pattern, -1, PREG_SPLIT_DELIM_CAPTURE);
```
milo: Why is it split it by single `%` and not by `%[^%]+%`?

src/Framework/Assert.php
```diff
@@ -509,6 +518,67 @@ public static function isMatching($pattern, $actual)
...
+		for ($i = count($parts); $i >= 0; $i--) {
+			$patternX = implode('', array_slice($parts, 0, $i));
```
milo: No need for 1st arg.

src/Framework/Assert.php
```diff
@@ -509,6 +518,67 @@ public static function isMatching($pattern, $actual)
...
+			$patternY = "$patternX%A?%";
+			if (self::isMatching($patternY, $actual)) {
+				$patternZ = implode('', array_slice($parts, $i));
+				break;
+			}
+		}
+
+		foreach (['%A%', '%A?%'] as $evilPattern) {
```
milo: I would name it hungryPattern :)

src/Framework/Assert.php
```diff
@@ -555,4 +625,14 @@ private static function isEqual($expected, $actual, $level = 0, $objects = NULL)
 		return $expected === $actual;
 	}
 
+
+	/**
+	 * @param  string $pattern
```
milo: No `$pattern` for consistency.

milo:
@JanTvrdik I left few commens in diff. Maybe squash the last two commits and I'm :+1: to merge.

milo:
@JanTvrdik May I ask for the state of this?

JanTvrdik:
Will try to finish this today.

src/Framework/Assert.php
JanTvrdik: I've renamed it o greedyPattern, is that ok?

src/Framework/Assert.php
JanTvrdik: I'm not sure it would work in all cases.

JanTvrdik:
@milo rebased and updated

milo:
Thank you! This is really great feature.

PR #315, opened 2016-05-18
Control environment variables setting refactoring

milo:
IMHO, runner's Job functionality should not depend on environment.

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -59,6 +59,11 @@ class CliTester
 		}
 
 		$runner = $this->createRunner();
+		$runner->setEnvironmentVariable(Environment::RUNNER, 1);
+		$runner->setEnvironmentVariable(Environment::COLORS, (int) Environment::$useColors);
+		if (isset($coverageFile)) {
+			$runner->setEnvironmentVariable(Environment::COVERAGE, $coverageFile);
+		}
 
 		if ($this->options['-o'] !== NULL) {
 			ob_clean();
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -77,7 +77,7 @@ class CliTester
 
 		$result = $runner->run();
 
-		if (isset($coverageFile)) {
+		if (isset($coverageFile) && preg_match('#\.(?:html?|xml)\z#', $coverageFile)) {
 			$this->finishCodeCoverage($coverageFile);
 		}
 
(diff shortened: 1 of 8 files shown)


dg:
:+1:

dg:
Passing it directly to proc_open didnt work, yes?

milo:
AFAIK, there was a lot of problems with it. Some extensions or even PHP behaved strangely with missing some system env vars. And it was hard to merge with `$_ENV`, because CGI has env vars in `$_SYSTEM`., IIRC.

But maybe it will work with PHP 5.4. I'll try it.

dg:
- setEnvVariable
- setEnvironmentVariable
- setEnvironmentVar

This method will nobody call, so I prefer `Environent` (already used in Tester) than `Env`.

milo:
Note: `$_ENV` content [depends](http://php.net/manual/en/reserved.variables.environment.php#98113) on INI `variables_order`. So it's affected by `-n`. Tested, it's true.

milo:
> I more prefer Environent

Renamed to `setEnvironmentVariable()`. I guess, private member names `envVars` are OK.

### According to `$_ENV`
With a simple script you can try, that `$_ENV` content depends on `variables_order` INI value:
```
php -n -r 'var_dump($_ENV);'  # PHP's default is EGPCS
It works.

php -r 'var_dump($_ENV);'  # I have GPCS in php.ini (PHP 5.6 on Windows, Debian too)
I does not work, $_ENV is an empty array.

php -d variables_order=EGPCS -r 'var_dump($_ENV);'
It works.
```

When I simply pass `$this->envVars + $_ENV` in [Job in here](https://github.com/nette/tester/blob/9cad20edca767717c80544b3e3e62f752d472558/src/Runner/Job.php#L95), tests [are failing](https://travis-ci.org/nette/tester/jobs/131116538).

When I update `.travis.yml` to run run tests like `php -d variables_order=EGPCS src/tester.php ...`, tests [pass](https://travis-ci.org/nette/tester/builds/131147521).

It looks it works (still, I'm not 100% sure).

### How to replace `$_ENV`
1) By `$_SERVER`. It depends on `variables_order` too, but IMHO it is set everywhere. But it will pass new env vars, because $_ENV is only $_SERVER's subset.

2) Obtain `$_ENV` content during [info.php](https://github.com/nette/tester/blob/9cad20edca767717c80544b3e3e62f752d472558/src/Runner/PhpInterpreter.php#L57) call

3) Use `phpinfo(INFO_ENVIRONMENT);` and parse output (it is HTML in CGI).

4) By some `system('set');` command.

5) Any other idea?


IMHO, I would stay with proposed set/unset env var by `putenv()`.

dg:
> I more prefer Environent

…and I cannot write it correctly :-)

Prvt vars are ok, of couse.

dg:
5) try to call getenv() on every $_SERVER keys?

milo:
@dg Good idea! I'll try it. Stanalone commit will be better for easy reverting :o)

milo:
Continues in #316

Issue #301, opened 2016-03-22
Provide thread number when running paralel


milo:
When run in paralel, there can be handy to have a current thread number. For example, prepare the same 8 databases, one for every thread.

Idea by @matej21

JanTvrdik:
Nette tester does not use threads. Do you mean PID?

hrach:
Probably not process id, but process number.

JanTvrdik:
I think I got it now but both thread number and process number are confusing terms.

milo:
In the doc is `-j 8` ... `Nette Tester uses 8 threads...`. I think the more precise terminology is queue.

fprochazka:
That is a great idea! I was always hacking it somehow and it was very fragile.

PR #317, opened 2016-05-23
Pass "thread" number to test

milo:
Solves #301. Not sure about:
- thread terminology - maybe introduce naming `queue`, `job queue`? Something better?
- `Job::getEnvironmentVariable()` - seems that Runner should not depend on it


dg:
„Thread“ is not perfect, but is good enough.

PR #319, opened 2016-06-17
Collector: fix save method on strict types

Mikulas:
Calling `Collector::save()` without Tester coverage enabled triggers
strict types warning, which is caught by tester environment and stops
test execution. This fix halts `save()` method before the callback is
checked.

Exception is intentionally not thrown as `save()` handling would be overly complex in user created files.

### DIFF
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -66,6 +66,10 @@ class Collector
 	 */
 	public static function save()
 	{
+		if (!self::isStarted()) {
+			return;
+		}
+
 		list($positive, $negative) = call_user_func([__CLASS__, self::$collector]);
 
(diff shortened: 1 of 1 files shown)


Mikulas:
This could be obviously fixed in userland with `if (!self::isStarted()) { … }`, but it's hard to debug in the first place. Explicitly checking in tester is more developer friendly. Are there reasons for not including this I'm overlooking?

milo:
There is no reason, I do not have a much time for Tester now.

Why do you call `Collector::save()` manually?

Mikulas:
#322 is problematic: wrapping all calls to `save()` with `if isStarted()` is exactly what I didn't want :)

Either solution is pretty good though, the exception in save() at least immediately makes clear what the problem is.

PR #328, opened 2016-07-01
in Linux environment variables are case-sensitive and uppercase [Closes #324]

### LINKED ISSUE #324: Environment variable TERM
In Fedora I had to change the used environment variable from **term** to **TERM**. Otherwise colors are not used in my tests.
Would it be possible to update the file _src/Framework/Environment.php_?
Thanks a lot.

Beutlin:


milo:
Thank you. Merged 97b0f798aba52fae036c84752e1d5f62ee981eb0

PR #329, opened 2016-07-07
The -o option can be used multiple times No.2

milo:
Alternative approach to #310.

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -29,6 +29,9 @@ class CliTester
 	/** @var bool */
 	private $debugMode = true;
 
+	/** @var string|null */
+	private $stdoutFormat;
+
 
 	public function run(): ?int
 	{
@@ -41,7 +44,7 @@ class CliTester
 		$this->debugMode = (bool) $this->options['--debug'];
 		if (isset($this->options['--colors'])) {
 			Environment::$useColors = (bool) $this->options['--colors'];
-		} elseif (in_array($this->options['-o'], ['tap', 'junit'], true)) {
+		} elseif (in_array($this->stdoutFormat, ['tap', 'junit'], true)) {
 			Environment::$useColors = false;
 		}
 
@@ -67,7 +70,7 @@ class CliTester
 			$coverageFile = $this->prepareCodeCoverage($runner);
 		}
 
-		if ($this->options['-o'] !== null) {
+		if ($this->stdoutFormat !== null) {
 			ob_clean();
 		}
 		ob_end_flush();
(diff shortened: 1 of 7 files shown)


dg:
`stdout` -> `stdoutFormat` :+1: 

`options['-o']` -> `$this->outputs` hmmm, perhaps unnecessary internal dependency…

I compared it on GitHub in phone and it is quite possible that I missed the other differences.

src/Runner/CliTester.php
```diff
@@ -101,21 +107,23 @@ private function loadOptions()
...
+    -c <path>                Look for php.ini file (or look in directory) <path>.
+    -l | --log <path>        Write log to file <path>.
+    -d <key=value>...        Define INI entry 'key' with value 'val'.
+    -s                       Show information about skipped tests.
+    --stop-on-fail           Stop execution upon the first failure.
+    -j <num>                 Run <num> jobs in parallel (default: 8).
+    -o <console|tap|junit|log|none>
+                             Specify one or more output formats with optional file name.
+                             (e.g. -o junit:output.xml)
+    -w | --watch <path>      Watch directory.
```
dg: What about `-o <console|tap|junit|log|none>  (e.g. -o junit:output.xml)` to save one row?

milo:
> options['-o'] -> $this->outputs hmmm, perhaps unnecessary internal dependency…

Yes. That's part I don't like too but I cannot figure how to solve it in other way. Imagine I will call `-o console -o tap:output.tap -o junit:output.xml`, I would like to:
- keep ENUM check in CommandLine. So value `tap:file.name` has to be modified before this check (NORMALIZER do it) and `file.name` has to be kept somewhere for OutputHandler's constructor. The `$this->outputs` is used for this. I cannot figure how to keep `file.name` inside `$options['-o']` without hidden functionality.

- prevent more outputs into one file

src/Runner/CliTester.php
milo: I'll try that.

milo: IMHO, it looks strange when not aligned.

milo:
@dg Updated

TomasVotruba:
What needs to be done here?

milo:
Not sure. But I'm not satisfied with an implementation.

milo:
That's what I'm nor sure about. The part with `NORMALIZER` probably. The main purpose for this PR is to allow pass multiple output handlers by CLI arguments. Propose is `-o format:file.ext`. Maybe another syntax would be better and more straightfoward for implementation.

PR #330, opened 2016-07-14
Option -C to keep system php.ini

milo:
@dg I'm aware of your propose of `-c default` but `-C` seems less magic to me. Anyway, if you think that saving one new option is better, I'm ok with it (maybe as `-c system`). See #331

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -103,6 +103,7 @@ Usage:
 Options:
     -p <path>                    Specify PHP interpreter to run (default: php).
     -c <path>                    Look for php.ini file (or look in directory) <path>.
+    -C                           Use system-wide php.ini.
     -l | --log <path>            Write log to file <path>.
     -d <key=value>...            Define INI entry 'key' with value 'val'.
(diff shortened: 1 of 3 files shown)


src/Runner/CliTester.php
```diff
@@ -102,6 +102,7 @@ private function loadOptions()
 
 Options:
     -p <path>                    Specify PHP interpreter to run (default: php).
+    -C                           Do not disable system php.ini.
```
dg: `Use system php.ini`? Or `default`?

src/Runner/PhpInterpreter.php
```diff
@@ -28,7 +28,7 @@ class PhpInterpreter
 	private $error;
 
 
-	public function __construct($path, array $args = [])
+	public function __construct($path, $keepSystemIni = FALSE, array $args = [])
```
dg: When all arguments are passed low-level in `$args`, the  `-n` should be part of `$args` too, or not?

dg:
I am fine with `-C`.

milo:
@dg Thanks for review! Updated.

Issue #326, opened 2016-07-01
The OutputHandler refactoring


milo:
@jiripudil wrote a PhpStorm pluging for Tester. I would like to refactor the `OutputHander` to be able better provide information for TeamCity output (according to https://github.com/jiripudil/tester/tree/feature/teamcity-output).

It will be with BC breaks I guess. My idea is to create class like:

``` php
class TestRunHistory
    public $file;
    public $stdout;
    public $exitCode;
    public $result;
    ...
}
```

and pass it to `OutputHandler` interface. But I didn't even try that concept.

milo:
Closing in favour of #370

PR #334, opened 2016-09-11
OutputHandler refactoring

jiripudil:
A take on the solution of #326. The basic idea is to give `OutputHandler`s more structured information about the tests. This involves creating a `TestInstance` object for each test (in other words, for each dot in the output); this object can either be instantiated with a `Job` object that is enqueued, or directly with results if the tests are marked as skipped or failed in the initiate phase. The output of these results is deferred to the main `Runner` loop in preparation for additional refactoring in #335.

There are BC breaks, but first, they affect _de facto_ internal code, and second, the issue targets the 2.0 milestone.

TODO
- [ ] Tests. I'd like to get some opinions first before I add them


src/Runner/Job.php
```diff
@@ -245,4 +245,24 @@ public function getHeaders()
...
 
+
+	/**
+	 * @return string
+	 */
+	private function buildArgumentsList()
+	{
+		$args = [];
+
+		foreach ($this->args as $key => $arg) {
```
jiripudil: Potential BC break: Job's arguments are now held in an associative array to provide more semantics. This implementation effectively disallows multi-valued options unless they are passed in the current fashion. Could this be a problem?

dg: IHMO current implementation is better, associative array disallows multi values as you said and --$key=$arg is not standard format.

jiripudil: What do you mean by "--$key=$arg is not standard format"? It's the exact same format as the one [used in the current implementation](https://github.com/nette/tester/blob/7c21d99/src/Runner/TestHandler.php#L54)

src/Runner/TestInstance.php
```diff
@@ -0,0 +1,161 @@
...
+ * Copyright (c) 2009 David Grudl (https://davidgrudl.com)
+ */
+
+namespace Tester\Runner;
+
+
+/**
+ * Carries information about a single test instance.
+ */
+class TestInstance
```
milo: IMHO a `Test` name would be sufficient.

milo: What about constants `Runner::FAILED, ::PASSED, ::SKIPPED`? This is maybe better place for them.

src/Runner/TestInstance.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		if (isset($args['multiple'])) {
+			$instanceNameParts[] = '#' . $args['multiple'];
+		}
+
+		if (isset($args['dataprovider'])) {
+			$instanceNameParts[] = '(' . $args['dataprovider'] . ')';
+		}
+
```
milo: These `method`, `multiple`, `dataprovider` references smell to me a little bit. Not sure how to do it. Maybe the `TestInstance` is now better place for homehow pass them to job.

src/Runner/Job.php
milo: @jiripudil The `key=val` is the Tester's "internal" format. Someone else can append own args from outside.

src/Runner/TestInstance.php
```diff
@@ -0,0 +1,161 @@
...
+		$this->result = $result;
+		$this->message = $message;
+		return $this;
+	}
+
+
+	/**
+	 * @return Job|NULL
+	 */
+	public function getJob()
```
milo: Not sure the `Job` should be part of an class interface.

src/Runner/TestInstance.php
```diff
@@ -0,0 +1,161 @@
...
+	private $instanceName;
+
+	/** @var int */
+	private $result;
+
+	/** @var string|NULL */
+	private $message;
+
+
+	private function __construct()
```
milo: Not necessary.

jiripudil:
Closing in favor of #345

PR #335, opened 2016-09-11
TeamCity output handler

jiripudil:
TeamCity is a CI output format used by JetBrains's line of products and is required by the [Tester plugin for PhpStorm](https://github.com/jiripudil/intellij-nette-tester)

Building on #334, this PR introduces one more breaking change: `OutputHandler::begin()` receives a list of `TestInstance`s so that it can analyze it and use the gathered information. `TeamCityPrinter` makes use of it to correctly open and close test suites / test cases.

Again, I'd like to add some tests once we come to an agreement on the implementation :)


Teyras:
Could you please elaborate on how to use the PHPStorm plugin correctly (with the `--setup` script)?

jiripudil:
@Teyras It was more of a note-to-self :) I'm working on a new release of the plugin that will have its own output handler bundled in the package, so relying on this specific branch of this specific fork will no longer be necessary

PR #342, opened 2016-12-02
Allow @skip on test methods.

### LINKED ISSUE #179: Skip one particular test
I know there is a way to skip a whole file of tests using anotation like this:

``` php
/**
 * @skip
 */
class Test extends TestCase
```

However, is there a way to skip one particular test in a file with more tests? Like so:

``` php
class Test extends TestCase
{
  public function testA()
  {
    // assert something
  }

  /**
   * @skip
   */
  public function testB()
  {
     // this is ran (regardless the anotation)
  }
}
```

Fuco1:
Right now users can only `@skip` entire file or `TestCase` class.  This patch implements a simple and consistent way to skip individual tests.

What you do is you add `@skip` annotation before the `testXYZ` method to skip it.  If the file has `@testCase` annotation then individual tests are skipped.  If it has none, the entire file is skipped (same as adding `@skip` on the class instead of method).

The assumption here is that if the test is *not* a `@testCase` skipping one test doesn't make sense as they are in some way related or dependent and represent a single "feature" which should be tested in entirety.

This fixes #179

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -78,6 +78,10 @@ class TestCase
 
 		$info = Helpers::parseDocComment($method->getDocComment()) + ['dataprovider' => NULL, 'throws' => NULL];
 
+		if (isset($info['skip'])) {
+			\Tester\Environment::skip($info['skip']);
+		}
+
 		if ($info['throws'] === '') {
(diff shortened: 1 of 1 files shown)


fmasa:
I think that skipping the whole file with `@skip` annotation on method brings WTF moment (or succesfull result for test suite with broken test). Wouldn't failing with message about requirement of @testCase annotation be better?

PR #345, opened 2016-12-19
Output handler refactoring

milo:
This is how I image solution for #326. So, the main purpose is to provide more information for OutputHandler.

This is a WIP concept with many TODOs to solve. One problem is, that the new `Test` class shares many common properties with `Job` and I'm not sure if it's OK.

### DIFF
--- a/src/Runner/Runner.php
+++ b/src/Runner/Runner.php
@@ -186,7 +186,7 @@ class Runner
 	{
 		$this->results[$test->getResult()]++;
 		foreach ($this->outputHandlers as $handler) {
-			$handler->result($test->getName(30), $test->getResult(), $test->message);
+			$handler->result(clone $test);
 		}
 
 		if ($this->stopOnFail && $test->getResult() === Test::FAILED) {
--- a/src/Runner/Output/ConsolePrinter.php
+++ b/src/Runner/Output/ConsolePrinter.php
@@ -47,19 +47,20 @@ class ConsolePrinter implements Tester\Runner\OutputHandler
 	}
 
 
-	public function result($testName, $result, $message)
+	public function result(Test $test)
 	{
 		$outputs = [
 			Test::PASSED => '.',
 			Test::SKIPPED => 's',
 			Test::FAILED => Dumper::color('white/red', 'F'),
 		];
-		echo $outputs[$result];
+		echo $outputs[$test->getResult()];
 
-		$message = '   ' . str_replace("\n", "\n   ", trim($message)) . "\n\n";
-		if ($result === Test::FAILED) {
+		$testName = $test->getName(30);
+		$message = '   ' . str_replace("\n", "\n   ", trim($test->message)) . "\n\n";
+		if ($test->getResult() === Test::FAILED) {
 			$this->buffer .= Dumper::color('red', "-- FAILED: $testName") . "\n$message";
-		} elseif ($result === Test::SKIPPED && $this->displaySkipped) {
+		} elseif ($test->getResult() === Test::SKIPPED && $this->displaySkipped) {
 			$this->buffer .= "-- Skipped: $testName\n$message";
 		}
 	}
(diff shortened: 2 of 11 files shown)


src/Runner/Test.php
```diff
@@ -0,0 +1,138 @@
...
+
+
+	/**
+	 * @param  int
+	 * @param  string
+	 * @return static
+	 */
+	public function withResult($result, $message)
+	{
+		# TODO: Vyjimku, pokud je test failed?
```
jiripudil: I'd throw an exception if the test already has *any* result

src/Runner/Test.php
```diff
@@ -0,0 +1,138 @@
...
+namespace Tester\Runner;
+
+
+/**
+ * Test represents one result.
+ */
+class Test
+{
+	const
+		NO_RESULT = 0, # TODO: lepsi jmeno
```
jiripudil: what about `PENDING`?

src/Runner/Runner.php
```diff
@@ -95,7 +90,7 @@ public function run()
 			$handler->begin();
```
jiripudil: I'd love to be able to get `$this->jobCount` in here, but that would mean moving this call after `$this->findTests()` in which results may already be sent to output. They'd have to be buffered somehow. What do you think?

src/Runner/Test.php
```diff
@@ -0,0 +1,138 @@
...
+	public function getArguments()
+	{
+		return $this->args;
+	}
+
+
+	/**
+	 * @return array
+	 */
+	public function getJobArguments()  # TODO: tohle sem asi nepatri
```
jiripudil: Now that the arguments are abstracted in an array, maybe `Job` should just call `getArguments()` and do this formatting itself? It feels coupled to the way Jobs are executed.

src/Runner/Runner.php
jiripudil: More precisely, the most perfect scenario would be if I could get an array of all `Test`s so that the output handler could process them (similarly to what I did in #335). It would make sense to me if `findTests` created and actually returned found `Test`s, and the `while` loop iterated over them instead of `Job`s. But that would mean that `Test`s would somehow have to be aware of their respective `Job`s, which is something you [didn't quite like](https://github.com/nette/tester/pull/334/files#r78320336)

jiripudil: I tried to outline one possible (and sensible to me) solution in milo/tester#1

jkuchar:
In original #334 there was `Test` which contains `Job`, here it is in reverse. What is the reasoning behind that? #345 feels more natural from test executor perspective. From user perspective original approach from #334 feels more natural. Is that what you had on mind?

milo:
The first part is merged a34637b01b2f86cc9830829c29b6e66411455138, 65ea0f85b4ed42481cd69c9d763303cfffd6ba00.

So, this PR is about OutputHandler only now. I'm thinking about BC. I have no idea, how many custom OutputHandlers exist. Maybe introduce a new interface.

milo:
@jiripudil Thanks for review.
@jkuchar From my point of view, is better that the Test is only independent wrapper and Job fills it.

jiripudil:
@milo You're welcome, sorry it took me so long.

What do you think of the idea behind milo/tester#1 ? It makes sense to me because some output handlers might use the information about the tests before printing the results.

I know it's a BC break, as is the rest of this pull, but since it targets 2.0, I think it's justifiable. Although a lot of the internal code has changed, the changes to the public API (`OutputHandler`) are minor and easy to follow.

I know @mrtnzlml has written a custom handler. As for others, maybe you could ask on Twitter?

jiripudil:
btw it seems that the `OutputHandler` API has already (potentially) broken BC by moving the result constants from `Runner` to `Test`, see jiripudil/intellij-nette-tester#30 🤔

milo:
Closing in favour of #370

PR #346, opened 2017-01-08
Runner: run *.phpt and *Test.php as default

milo:
- bug fix? no
- new feature? yes #278
- BC break? yes
- doc PR: (will)

Not sure for what was the `Runner::TEST_FILE_EXTENSION` good for.

### DIFF
--- a/src/Framework/Dumper.php
+++ b/src/Framework/Dumper.php
@@ -347,7 +347,7 @@ class Dumper
 	 */
 	public static function saveOutput($testFile, $content, $suffix = '')
 	{
-		$path = self::$dumpDir . DIRECTORY_SEPARATOR . basename($testFile, '.phpt') . $suffix;
+		$path = self::$dumpDir . DIRECTORY_SEPARATOR . pathinfo($testFile, PATHINFO_FILENAME) . $suffix;
 		if (!preg_match('#/|\w:#A', self::$dumpDir)) {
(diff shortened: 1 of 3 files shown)


klapuch:
@milo What if my test folder includes PHPUnit tests (because of snapshot testing) and Nette Tester tests? Nowadays, I have to list all my folders with Nette Tester tests and pass them to CLI command. In the old version I just passed `Tests/` folder and I was fine. Can I somehow force the old rule with `*.phpt`?

Thanks

JanTvrdik:
> I have to list all my folders with Nette Tester tests and pass them to CLI command

That seems like something you should have always be doing.

milo:
@klapuch You can reorganize your test files, e.g.
```
tests\
    Tester\
    PHPUnit\
```
Or, if your shell supports it, use `**` wild card, e.g. `vendor/bin/tester  tests/**/*.phpt`.

klapuch:
@JanTvrdik Why? Using Nette Tester to recursively execute all my tests in `Tests/` folder was always fine for me and I had no problems. When I create brand new folder for my tests (in `Tests/`), I have to put it to `build.xml`. It is very easy to forget it.

klapuch:
@milo Yeah, I though about the structure too, but honestly, I would like to keep my tests independently on framework. Does the `**` wildcard is supported by default when I run tester via `vendor/bin/tester`?

milo:
Function of `**` depends on your shell, not on the Tester.

PR #347, opened 2017-01-08
TestCase: ensure that parents tearDown() is called

milo:
- bug fix? no
- new feature? yes
- BC break? yes
- doc PR: will

Not sure it is a good idea. People use the `tearDown()` for assertions. It can be accidentally skipped by inheritance.

Or maybe add new method(s), e.g. `preAssert()`, `postAssert()`, and check it there.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -24,6 +24,9 @@ class TestCase
 	/** @var callable|NULL|FALSE */
 	private $prevErrorHandler = FALSE;
 
+	/** @var bool */
+	private $tearDownCalled;
+
 
 	/**
 	 * Runs the test case.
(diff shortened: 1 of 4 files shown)


src/Framework/TestCase.php
```diff
@@ -24,6 +24,9 @@ class TestCase
 	/** @var callable|NULL|FALSE */
 	private $prevErrorHandler = FALSE;
 
+	/** @var bool */
+	private $toreDown;
```
JanTvrdik: torn down?

milo: Typo. Hm, tearDownCalled will be better.

f3l1x:
I think it might be useful. At least it requires to call that method.

milo:
Looks that no one of `TestCase` users care, so getting out from milestone 2.0.

Issue #275, opened 2015-11-24
instruction "--bootstrap" like PHPUnit


Darracq:
Maybe it would be a good idea to implement the instruction "--bootstrap" like PHPUnit

```
phpunit --bootstrap src/autoload.php tests/MoneyTest
```
(from https://phpunit.de/getting-started.html)

With this we avoid writing in each *.phpt file lines like:
```
require __DIR__ . '/bootstrap.php';
//  if you want have the same folder structure like your '/app':
require __DIR__ . '/../../../../../bootstrap.php';
```

It also allows you to use different "bootstrap.php" if you need testing in different environments.
```
tester --bootstrap 'bootstrap_test_server.php'
tester --bootstrap 'bootstrap_local_server.php'
```

milo:
Tester is different. Every test behaves as an ordinary PHP file and you can run it as:
```php
php tests/dir/test.phpt
```
Or you can easily debug it and step through it IDE.

`vendor/bin/tester` is only tests runner. It takes every single test and run it as a standalone process.

When you consider this, some kind of bootstrapping is necessary.

I have one workaround on my mind... you can try
`vendor/bin/tester -d include_path=/dir/with/bootstrap tests`

Basic tutorial is at https://tester.nette.org/

milo:
And next ugly workaroung... set environment variable and use it in tests
```
BOOT=/path/to/bootstrap.php vendor/bin/tester tests
```

milo:
I realized, this feature is already implementeed by PHP itself:
```
vendor/bin/tester -d auto_prepend_file=/path/to/bootstrap.php tests
```

Fuco1:
I would much rather not use php language flags, they are pretty unreliable, depend on ini settings and god knows what else. Having this feature is as simple as 

```php
		$autoloadFile = $input->getOption('autoload-file');
		if ($autoloadFile !== null && is_file($autoloadFile)) {
			$autoloadFile = $fileHelper->normalizePath($autoloadFile);
			if (is_file($autoloadFile)) {
				require_once $autoloadFile;
			}
		}
``` 

(code from [phpstan](https://github.com/phpstan/phpstan/blob/master/src/Command/AnalyseCommand.php#L69))

Is there really any drawback of having such an option?  It would save a lot of useless `require_once` in tests.

milo:
No drawbacks. I just don't see any advantage.

Fuco1:
[...]

PR #355, opened 2017-04-17
Execute provided script before every test (`--bootstrap`)

jkuchar:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: 

There should be some global way how to configure environment of tests. Yep I'm looking at you `Environment::setup()`. 

- `--bootstrap` script should configure only environment only in way that tests can run without it
- `Environment::setup()` is probably good candidate for bootstrap script (at least coverage setup part)
- In future this can be used to simplify running coverage / setting up custom coverage collector (which needs to be done inside test but has no value for test itself)
- In contract putting auto-loader there is **very bad idea** as tests are not standalone anymore
- I have added `stdout` to `PhpInterpreter` exception when execution error occurs; this simplifies debugging if bootstrap fails for some reason

Discussion: is `--bootstrap` sufficient name? In real it should not bootstrap test, it should bootstrap test environment for Tester. It should have (almost) no connection to test itself.

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -54,6 +54,11 @@ class CliTester
 			return;
 		}
 
+		if (isset($this->options['--bootstrap'])) {
+			$exportedPath = var_export($this->options['--bootstrap'], TRUE);
+			$this->interpreter->addPhpIniOption('auto_prepend_file', $exportedPath);
+		}
+
 		if ($this->options['--coverage']) {
 			$coverageFile = $this->prepareCodeCoverage();
 		}
@@ -112,7 +117,8 @@ Options:
     -o <console|tap|junit|none>  Specify output format.
     -w | --watch <path>          Watch directory.
     -i | --info                  Show tests environment info and exit.
-    --setup <path>               Script for runner setup.
+    --setup <path>               Script for test runner setup. (configure Tester)
+    --bootstrap <path>           Script executed before each test (configure test environment)
     --colors [1|0]               Enable or disable colors.
     --coverage <path>            Generate code coverage report to file.
     --coverage-src <path>        Path to source code.
@@ -123,6 +129,7 @@ XX
 			'-c' => [CommandLine::REALPATH => TRUE],
 			'--watch' => [CommandLine::REPEATABLE => TRUE, CommandLine::REALPATH => TRUE],
 			'--setup' => [CommandLine::REALPATH => TRUE],
+			'--bootstrap' => [CommandLine::REALPATH => TRUE],
 			'paths' => [CommandLine::REPEATABLE => TRUE, CommandLine::VALUE => getcwd()],
 			'--debug' => [],
 			'--coverage-src' => [CommandLine::REALPATH => TRUE],
--- a/src/Runner/PhpInterpreter.php
+++ b/src/Runner/PhpInterpreter.php
@@ -64,7 +64,7 @@ class PhpInterpreter
 		$output = stream_get_contents($pipes[1]);
 		$this->error = trim(stream_get_contents($pipes[2]));
 		if (proc_close($proc)) {
-			throw new \Exception("Unable to run $path: " . preg_replace('#[\r\n ]+#', ' ', $this->error));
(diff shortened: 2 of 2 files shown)


src/Runner/CliTester.php
```diff
@@ -54,6 +54,11 @@ public function run()
 			return;
 		}
 
+		if (isset($this->options['--bootstrap'])) {
+			$exportedPath = var_export($this->options['--bootstrap'], TRUE);
+			$this->interpreter->addPhpIniOption('auto_prepend_file', $exportedPath);
```
jkuchar: This should work also with HHVM [test](https://github.com/facebook/hhvm/blob/46a9a688340dd4f38806c4ca1e5798c13fb5264c/hphp/test/slow/program_functions/auto_prepend_file-1.php.ini)

dg:
You can use `-d auto_prepend_file=file`

jkuchar:
You are right, that is how it works anyway.

So how do I get to this idea...

I want to make coverage collector more versatile, e.g. to support branch coverage or performace data collection, this needs to be set-up somewhere. And only place this can be set up is inside test's environment.

You can say, lets put it into bootstrap.php, which exists in every test suite by convention. And you are right. Technically you can do it. But this code has nothing to do with test logic itself, it is Tester internal infrastructure code? So I want to provide tool which seems legit (=has intuitive name), where you configure these things and do not mess your tests `bootstrap.php` with this tester-related infrastructure code.

Btw when you then want to switch coverage generator which depends on custom coverage collector (e.g. branch coverage collector), you should be able to do that just by providing configuration options to Tester CLI. Does that make sense?

One more idea - you can write completely Tester independent tests in Tester (e.g.`<?php if(doSth() === 5) exit(1);`). This test has no dependencies, and it rund inside Tester without any additional code. Until... Until you try to run coverage. Then Tester needs its infrastructure code to be in place for every test from test suite, otherwise coverage report will not include tests without coverage collector. (BTW there is currently no way how to detect these tests) Important to mention is that injected code must not affect result or execution of test, it just prepares PHP's environment to collect coverage, that's it.

I was considering to make `--bootstrap=autostart-coverage.php` as default value, which will just make sure that every test will have coverage code injected.

TL;DR: I want to make this configurable because I want to have easy to use injection point for configuring custom coverage collector.

jkuchar:
Maybe rename to `--coverage-collector=file`? This file will be included only when coverage is enabled in Tester and there can be default implementation which inits default coverage collector. This will allow Tester to run coverage on any test (even completely independent ones) and in future configure custom coverage collector.

Downside is that there is no way how `--coverage-collector` can configure environment to pretty print error messages as it is run ony when coverage is run. (e.g. to pretty print exception from tests that are completely Tester independent)

dg:
Related to #275

milo:
@jkuchar It is a long reading, I hope I understand.

You want to introduce some kind of setup tasks, so test will run like:
```php
require 'setup-task-1.php';
require 'setup-task-2.php';
require 'test-itself.phpt';
```

Besides the concept, there is a one technical problem. You can have only one `auto_prepend_file`. So multiple tasks must be multiplexed into single file.

According to `--bootstrap` option, I would prefer to use current `--setup`. Something like:
```php
$runner->addSetupScript('path/to/task-1.php');
$runner->addSetupScript('path/to/task-2.php');
```

Next question is, how to multiplex it into single script:
1) Create script in temporary directory. It is problematic, Tester does not use any temp dir. And not sure about `open_basedir` in CLI.
2) Use some script embeded into Tester, some `prepend.php` and pass scripts via environment:
```
TESTER_SETUP_SCRIPT_COUNT=2
TESTER_SETUP_SCRIPT_1=path/to/task-1.php
TESTER_SETUP_SCRIPT_2=path/to/task-1.php
```
3) ???

In any case, `auto_prepend_file` can be only one. So any mechanism using it must take user's or system's `auto_prepend_file` setting in to mind.

Besides all of this, functionality to run test as a simple script by `php test.phpt` has to be kept.

dg:
I'm closing PR, it's not a much requested feature.

PR #356, opened 2017-04-17
bootstrap and Environment::setup() should be idempotent

jkuchar:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: not really needed

I could not find any reason why including bootstrap twice or calling `Environment::setup()` twice would result into error. Any arguments against?

This can help with migration to `--bootstrap` option from `bootstrap.php` files where `Environment::setup()` is usually called. (related to #355)

### DIFF
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -44,6 +44,12 @@ class Environment
 	 */
 	public static function setup()
 	{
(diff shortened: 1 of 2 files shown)


milo:
From my point of view, `require_once` is a sign of bad design. Closing it since it is not needed for now. Can be reopend later.

Issue #218, opened 2015-03-20
@skipOthers


hrach:
This is similar to https://github.com/nette/tester/issues/179. Personally, I quite often need to disable other tests methods in testcase to test only one method. Adding @skip to other methods is demanding. So I would appriciate something like @skipOthers annotation, which would allow test only the specific testcase method. There are two reason why it's convenient:
- if test is run by browser / php, there is no test method sandboxing.
- if test has more methods and some sections, it would be faster to test it.

fprochazka:
Are you aware you can run just one testing method by running the file ?

```bash
$ php tests/MyTest.phpt testMethod
```

hrach:
This was the mistake which was presented on posobota, without having clue why it's working. This is undocumented internal fetarue, which don't work with bin/tester - that was the ackward part of talk when someone (don't know who) didn't know why it's not working. So I am aware, but it is not solution, for example it will not run test based on dataprovider, etc.

fprochazka:
I believe it will run tests even with data provider :)

hrach:
I was talking about file dataprovider: https://github.com/nette/tester/blob/a3139819bb013256c7fc64fea37070102dab2c30/Tester/Framework/Environment.php#L168 I would have to run `php test.phpt testMethod` with `--dataprovider` manually.

milo:
One possibility is `$testCase->run('testMethod');` calling.

hrach:
Good point! For now It's enough for me :)

hrach:
@milo It's not working!

```
Cannot list TestCase methods in file '...\relationships.oneHasMany.phpt'. Do you call TestCase::run() in it?
```

hrach:
The current solution does not solve|fix this issue. If I pass the method name as an argument, it runs the testcase only one and ignores @dataProvider.

milo:
Will be solved by https://github.com/nette/tester/commit/710c4b5957f95147a67b6d43f672cea9de545d61 where is a data provider applied.

dg:
So can we release new version?

hrach:
Just side note: in phpunit there is a cli --filter option, which allows you to filter method like 
`phpunit --filter '::testFoo' mytest.phpt`. It's pretty addictive. The advantage is that you don't need to update your test, you can reuse history in console, ...

milo:
@hrach `php path/to/test-case.phpt --method=testMe` will work as well

milo:
@dg #261

hrach:
[...]

PR #358, opened 2017-05-03
TestCase: implemented @skip and @skipOthers for test methods

milo:
- bug fix? no (#179, #218, #342)
- new feature? yes
- BC break? no
- doc PR: will

This PR adds support for `@skip` and `@skipOthers` for TestCase test methods. When `@skip` is used, method is skipped. When `@skipOthers` is used, method is run and all other methods are skipped.

This can be implemented in two ways. From `Runner` point of view (this implementation) or from `TestCase` point of view (the test is run and skip is done in runtime by `TestCase::runTest()`).

Both ways require `@testCase` to be used. Just to prevent that one `@skip` annotation skips silently all others test methods.

I choosed the first alternative because it saves resources.

TODO: Now, when `@testCase` is not used, the `@skip` and `@skipOthers` are silently ignored. So probably `TestCase::run()` should fail to indicate, that something can be wrong.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -15,7 +15,9 @@ class TestCase
 {
 	/** @internal */
 	const LIST_METHODS = 'nette-tester-list-methods',
-		METHOD_PATTERN = '#^test[A-Z0-9_]#';
+		METHOD_PATTERN = '#^test[A-Z0-9_]#',
+		METHOD_RUN = 'RUN',
+		METHOD_SKIP = 'SKIP';
 
 
 	/** @var bool */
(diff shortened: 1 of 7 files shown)


src/Framework/TestCase.php
```diff
@@ -35,23 +37,54 @@ public function run()
...
+		/** @var \ReflectionMethod[] $methods */
+		$methods = array_filter((new \ReflectionObject($this))->getMethods(), function (\ReflectionMethod $rm) {
+			return preg_match(self::METHOD_PATTERN, $rm->getName());
+		});
 
 		if (isset($_SERVER['argv']) && ($tmp = preg_filter('#--method=([\w-]+)$#Ai', '$1', $_SERVER['argv']))) {
 			$method = reset($tmp);
 			if ($method === self::LIST_METHODS) {
 				Environment::$checkAssertions = FALSE;
+
```
f3l1x: What about separate ti to own method?

Fuco1:
Any news on this one?  I would love to see this in tester to make it more competitive with other frameworks. (also I really quite need this :D)

milo:
At least, feedback is needed. I don't use TestCase to often and Nette itself the same. Thats the reason I didn't merge it.

Did you read related issues discussion, mentioned in the PR description? Maybe you can find another solution for your case there.

PR #362, opened 2017-06-29
Tests ordering

milo:
- bug fix? no
- new feature? yes
- BC break? no, but temporary directory is required
- doc PR: will

Runner remebers last run tests state and runs failed and new tests as first.

### DIFF
--- a/src/Runner/Runner.php
+++ b/src/Runner/Runner.php
@@ -48,6 +48,12 @@ class Runner
 	/** @var bool */
 	private $interrupted;
 
+	/** @var string|NULL */
+	private $tempDir;
+
+	/** @var array */
+	private $lastResults = [];
+
 
 	public function __construct(PhpInterpreter $interpreter)
 	{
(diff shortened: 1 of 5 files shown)


src/Runner/Runner.php
```diff
@@ -95,6 +101,27 @@ public function run()
 		}
 		$this->jobCount = count($this->jobs) + array_sum($this->results);
 
+		if ($this->tempDir) {
+			$this->orderFile = $this->tempDir . DIRECTORY_SEPARATOR . 'Tester.Runner-' . substr(md5(__FILE__), 0, 5) . '.list';
+			touch($this->orderFile);
```
dg: I usually run globally installed Runner, so maybe `getcwd()` should be in hash too, what you think?

dg:
GREAT!

src/Runner/Runner.php
milo: Probably yes. Will see how it will work in practice. Can by changed later to global, not hashing at all.

dg: This is actually true, there may not be hash.

src/Runner/Runner.php
```diff
@@ -21,6 +21,9 @@ class Runner
 	/** @var int  run in parallel threads */
 	public $threadCount = 1;
 
+	/** @var string */
```
JanTvrdik: `NULL|string`?

src/Runner/Runner.php
```diff
@@ -48,6 +51,9 @@ class Runner
 	/** @var bool */
 	private $interrupted;
 
+	/** @var string */
```
JanTvrdik: `NULL|string`

milo:
@dg @JanTvrdik What about such approach? (have to fix tests yet)

JanTvrdik:
@milo Looks awesome.

dg:
Separated files for each test instead of one file per tested folder? I am not sure what is better, but current code is more readable.

milo:
With separated files is much easier to update result of one test when change.

Reading 1 byte from 1000 files took 40ms on old server with rotation disks. I think performace will be OK.

PR #367, opened 2017-07-12
Dropped HHVM support

milo:
- bug fix? yes
- new feature? no
- BC break? yes


dg:
Bye bye HVMM, rest in peace :)

PR #368, opened 2017-07-21
Dropped PHP < 5.6 support

milo:
- BC break? yes

Any tips to use new features from [PHP 5.5](http://php.net/manual/en/migration55.new-features.php) or [PHP 5.6](http://php.net/manual/en/migration56.new-features.php)?

In short:
```
PHP 5.5
- generators and yield
- finally
- password_hash()
- foreach ($var as list(a, b)) {}
- empty(from expression)
- direct array and string literals access a = [a, b, c][0], 'str'[1]
- ::class
- opcache extension
- non-scalar keys type (from iterators) support in foreach()

PHP 5.6
- constant expressions - const A + 10
- variadic functions ...
- argument unpacking ...
- exponent opeartor **
- use const, use function
- phpdbg SAPI
- default charset encoding
- php://input is reusable
- 2GB+ files upload support
- GMP objects overload operators
- hash_equals()
- __debugInfo()
- gost-crypto hash algorithm


dg:
👍

PR #370, opened 2017-07-31
OutputHandler v3

milo:
- bug fix? no, related to #345
- new feature? yes
- BC break? **YES**
- doc PR: will

This BC break storm can be split on:
---
**OutputHandler receives whole Test instance**
So it can reach all its properties, including stout and stderr.

**Added `OutputHandler::prepare(Test $test)`**
So it can perform some prepare actions. AFAIK it is "needed" by @jiripudil for PhpStorm plugin, right?

<del>**OutputHandler is an abstract class instead of the interface**
OutputHandler is the only one place where someone can hook the Tester. It will help to prevent BC breaks</del>

**Dropped runner's `getResults()` and `getJobCount()`**
With `OutputHandler::prepare()` is easy to count jobs and results by its own. Runner itself does not need this methods.

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -201,10 +201,10 @@ XX
 		if ($this->options['-o'] !== 'none') {
 			switch ($this->options['-o']) {
 				case 'tap':
-					$runner->outputHandlers[] = new Output\TapPrinter($runner);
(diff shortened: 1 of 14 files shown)


milo:
Updated to keep `OutputHandler` as an interface.

Issue #379, opened 2017-11-28
Marking test as skipped will prevent execution of all subsequent tests


trebi:
- bug report? yes
- feature request? no
- version: 2.0.0 

### Description
If test is marked as skipped, all subsequent tests will not execute.

### Steps To Reproduce

1. Create file
```php
<?php declare(strict_types=1);

namespace Tester\BugReport;

use Tester\Assert;

require __DIR__ . '/../bootstrap.php';

/** @TestCase */
final class BugReportTest extends \Tester\TestCase
{
	public function testSkip(): void
	{
		\Tester\Environment::skip();
	}


	public function testFail(): void
	{
		Assert::true(false);
	}
}

(new BugReportTest)->run();
```
2. Run the file, the output will be:
```
Skipped:


Process finished with exit code 177
```
giving no indication that second test is failing. Such behaviour causes that I prefer commenting out the test rather than makring it as skipped.

BTW if you swap the order of the tests, it will execute the failing test and fail as expected, which is inconsistent behaviour. Order of methods should never matter.

milo:
If you run this test by Tester, you'll get two results: `sF` That's correct. One skipped, one failed. Every test method runs in own separated process.

But if you run it as `php test.phpt`, the `@TestCase` is ignored (as mentioned in doc), and all test methods run one by one in one process, so you get only one result.

You can try:
```
php test.phpt --method testSkip
php test.phpt --method testFail
```

trebi:
Thanks for reply, but I think it would be handy that even in "serial mode" it would run all the tests until the first real failure (skip is not really a failure), or you will be sure that all tests have passed.

milo:
I hope solved by #380

PR #380, opened 2017-11-28
Skip does not kill process when not in runner

trebi:
- bug fix? yes #379
- new feature? no
- BC break? probably not, tests pass without any change

To explain the situation where this helps:

When writing a class, you usually have another test class that holds your unit tests. You make a change in the class, quickly skip (CTRL+ALT+T) to the test class, press run (CTRL+ALT+F10) to make sure nothing got broken and move to another change.

In the current state, such workflow is impossible, because if you have a skipped test in the class, execution simply stops on that skipped test an no more test are executes, so you don't know if they are broken or not. You will note it once you run the whole suite, which is unnecessarily late. Current behavior  forces the programmer to comment-out tests rather than using SKIP, which makes SKIP an useless construct.

### DIFF
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -139,8 +139,11 @@ class Environment
 	public static function skip($message = '')
 	{
 		self::$checkAssertions = false;
-		echo "\nSkipped:\n$message\n";
-		die(Runner\Job::CODE_SKIP);
+
+		if (\getenv(self::RUNNER)) {
+			echo "\nSkipped:\n$message\n";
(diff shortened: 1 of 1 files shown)


milo:
Thanks, merged.

dg:
This is annoying. I was wondering why I manually run test doesn't pass, and it did not meet the conditions, but skip() does absolutely nothing…

milo:
And now I realized, that Runner may not be the only automated executor, so test must be skipped even when run by anything else.

PR #382, opened 2018-01-02
Extract storeAssertDiff from Dumper::dumpException

trejjam:
- bug fix? no
- new feature? yes
- BC break? no

This refactor Dumper::dumpException in the way that part maintaining storing Assert diff is callable from other test libraries.

### DIFF
--- a/src/Framework/Dumper.php
+++ b/src/Framework/Dumper.php
@@ -258,46 +258,11 @@ class Dumper
 		}
 
 		if ($e instanceof AssertException) {
-			$expected = $e->expected;
-			$actual = $e->actual;
-
-			if (is_object($expected) || is_array($expected) || (is_string($expected) && strlen($expected) > self::$maxLength)
-				|| is_object($actual) || is_array($actual) || (is_string($actual) && strlen($actual) > self::$maxLength)
-			) {
-				$args = isset($_SERVER['argv'][1])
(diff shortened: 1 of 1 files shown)


milo:
I'll check it during next week.

milo:
Btw what's the purpose is?

Trejjam:
I use it with [mango tester](https://github.com/mangoweb-backend/tester-infrastructure) to a store diff.

milo:
Thanks for rebase!

Your `storeAssertDiff()` does two things. Translates exception into message and creates dump files (arg reference is not nice public API). Do you need both?

I know mango flavoured Tester, but not its source code in detail. Could you elaborate more in detail why do you need extract this functionality? From your description it looks like you only need to remove `@internal` from `saveOutput()` method.

Trejjam:
The storing part is what I would like to port into Mango tester.

It uses own [TestCase](https://github.com/mangoweb-backend/tester-infrastructure/blob/master/src/TestCase.php#L166), where the only message about assert fault is reported, not simple fault then require xDebug or the other inspection method.

What about removing `@internal` or simplifying PR to move only logic in https://github.com/nette/tester/blob/master/src/Framework/Dumper.php#L264 and make it publicly callable?

You are right that storeAssertDiff API is not good enough.

PR #385, opened 2018-05-09
DomQuery: added missing error suppression for <picture>

JanTvrdik:
- bug fix? yes
- new feature? no
- BC break? no

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture


dg:
Thx

Issue #232, opened 2015-06-11
Multiple folders/paths for --coverage-src parameter


jeroneemou:
I thought it would be nice to insert multiple directories/paths in `--coverage-src` parameter to create coverage report upon. So far I can use only one folder.

milo:
Related to #225

milo:
Solved by #388

Issue #225, opened 2015-04-23
Coverage: folder skip option


f3l1x:
At this moment there are only 2 options.

```
--coverage FILE
```

```
--coverage-src FOLDER
```

I would like to skip some nested folder. Something like

```
--coverage-skip FOLDER/A
```

Is there any other way how to do that?

milo:
It's not possible now. Just curious... what is the `FOLDER/A` contents to be skipped?

f3l1x:
Sure, It's 3rd part library which is under our project namespace and there's no composer package.

f3l1x:
I think I can accomplish this issue with this commit https://github.com/nette/tester/commit/2286131c7bd9d47ea7fc638480d4ea1f2ace37aa. Thanks.

PR #388, opened 2018-09-05
Repeatable --coverage-src option

harmim:
- new feature
- BC break? yes

Added support for repeatable `--coverage-src` option. It is useful when you have got eg. multiple directories in root of your project which you want to generate coverage for.

### DIFF
--- a/src/CodeCoverage/Generators/AbstractGenerator.php
+++ b/src/CodeCoverage/Generators/AbstractGenerator.php
@@ -26,8 +26,8 @@ abstract class AbstractGenerator
 	/** @var array */
 	protected $data;
 
-	/** @var string */
-	protected $source;
+	/** @var array */
+	protected $sources;
 
 	/** @var int */
 	protected $totalSum = 0;
@@ -38,9 +38,9 @@ abstract class AbstractGenerator
 
 	/**
 	 * @param  string  $file  path to coverage.dat file
-	 * @param  string  $source  path to covered source file or directory
+	 * @param  array   $sources  paths to covered source files or directories
 	 */
-	public function __construct(string $file, string $source = null)
+	public function __construct(string $file, array $sources = [])
 	{
 		if (!is_file($file)) {
 			throw new \Exception("File '$file' is missing.");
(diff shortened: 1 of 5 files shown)


milo:
This PR looks great!
Solves #232, partially solves #225

harmim:
So is it ready to be merged?

milo:
Not yet. IMHO one thing should be done differently but I have to try it.

src/CodeCoverage/Generators/HtmlGenerator.php
```diff
@@ -95,8 +95,13 @@ private function parse(): void
 			}
 
 			$light = $total ? $total < 5 : count(file($entry)) < 50;
+			$nameReplacePairs = [];
```
milo: There should be single replacement, common path of all sources. Otherway duplicates may appear.
And it can be done before `foreach` loop.

src/CodeCoverage/Generators/AbstractGenerator.php
```diff
@@ -103,9 +108,13 @@ public function getCoveredPercent(): float
 
 	protected function getSourceIterator(): \Iterator
 	{
-		$iterator = is_dir($this->source)
-			? new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($this->source))
-			: new \ArrayIterator([new \SplFileInfo($this->source)]);
+		$iterator = new \AppendIterator();
```
milo: No brackets according to CS consistency.

src/CodeCoverage/Generators/AbstractGenerator.php
```diff
@@ -62,12 +62,17 @@ public function __construct(string $file, string $source = null)
 				}
 			}
 			$source = dirname($source . 'x');
+			$sources[] = $source;
```
milo:
```php
$sources = [dirname(...)];

File: src/CodeCoverage/Generators/AbstractGenerator.php [hunk 4]
```diff
@@ -103,9 +108,13 @@ public function getCoveredPercent(): float
...
-		$iterator = is_dir($this->source)
-			? new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($this->source))
-			: new \ArrayIterator([new \SplFileInfo($this->source)]);
+		$iterator = new \AppendIterator();
+		foreach ($this->sources as $source) {
+			$partialIterator = is_dir($source)
+				? new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($source))
+				: new \ArrayIterator([new \SplFileInfo($source)]);
+			$iterator->append($partialIterator);
+		}
```
milo: Since `$partialIterator` is not used, write it this way:
```php
$iterator->append(
    is_dir($source)
        ? new \Recursive.....,
        : new \Array.....;
);

milo:
Sorry if you see any mess here. I'm trying the review and resolve/unresolve conversation functionality.

src/CodeCoverage/Generators/HtmlGenerator.php
harmim: @milo What is the easiest way to find common path of all sources?

milo: [This code](https://github.com/nette/tester/blob/ca80e7bb1210c5e2141925b8f74ba7cf43500843/src/CodeCoverage/Generators/AbstractGenerator.php#L55-L63) can be reused.

milo:
Merged 2286131c7bd9d47ea7fc638480d4ea1f2ace37aa. Thank you!

PR #392, opened 2018-10-24
Helper::purge() protection [WIP]

dg:
- new feature?  
- BC break? no

1) Prevents an incorrect call with an empty parameter (eg due to a typo in the code), which can lead to the whole disc being deleted

2) Disables `purge('/')` or `purge('c:/') just for sure

### DIFF
--- a/src/Framework/Helpers.php
+++ b/src/Framework/Helpers.php
@@ -20,6 +20,9 @@ class Helpers
 	 */
 	public static function purge(string $dir): void
 	{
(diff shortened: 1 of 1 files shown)


milo:
Hope you didn't discover that on your own PC.

Not sure there should be a test for it :)

milo:
Maybe run this check after `realpath(dirname($dir))` to prevent `.` or `..` misusing too.

PR #396, opened 2019-02-02
Add PCOV support

trejjam:
- new feature
- BC break? no

Hello,

I met an announcement about PHPUnit 8 support [PCOV](https://github.com/krakjoe/pcov) which should be really fast in comparison to other code coverage possibilities, e.g. https://twitter.com/kornrunner/status/1091381371447230464, https://twitter.com/RemiCollet/status/1091336903423594496, etc.

This PR adds support for it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "nette/tester",
 	"description": "Nette Tester: enjoyable unit testing in PHP with code coverage reporter. 🍏🍏🍎🍏",
 	"homepage": "https://tester.nette.org",
-	"keywords": ["testing", "unit", "nette", "phpunit", "code coverage", "xdebug", "phpdbg", "clover", "assertions"],
+	"keywords": ["testing", "unit", "nette", "phpunit", "code coverage", "xdebug", "phpdbg", "pcov", "clover", "assertions"],
 	"license": ["BSD-3-Clause", "GPL-2.0", "GPL-3.0"],
 	"authors": [
 		{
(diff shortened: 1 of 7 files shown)


milo:
@trejjam Great PR! But I have to check, why I get 0% coverage with it :)

dg:
You should write more tests :)

trejjam:
To make it work, I needed to configure application path in [pcov.directory](https://github.com/krakjoe/pcov#configuration) for my environment (probably because my test entrypoint is in different file subtree then application - which is out of scope in default/auto configuration)

milo:
The `pcov.directory` should be set to a common path of all `--coverage-src` values.

The `pcov.directory` is `SYSTEM,PERDIR` so must be set by `-d pcov.directory=...` from [CliTester](https://github.com/nette/tester/blob/1e7351829190bd07db8cf7d5ac326937dec6a50c/src/Runner/CliTester.php#L225). Any idea how to do it nice and clean?

1) Add `PhpInterpreter::getCodeCoverageEngine(): ?string` and deprecate `PhpInterpreter::canMeasureCodeCoverage()`

2) ?

milo:
With #400 there is easy to detect PCOV and set `-d pcov.directory` INI value.

milo:
Btw. if theoretically, all code coverage engines (PHPDBG, PCOV, Xdebug) are present, which one would you prefer?

Compare https://milovo.cz/static/tester-396/

trejjam:
I see more points of view.
- prefer Xdebug - it can cover more edge cases - like method signature, and close brace (I feel it like more robust solution)
- prefer PCOV - it is fast, lightweight, not requires Xdebug (for PHP 7.3 is still not stable)

I will prefer a fixed order of engines. For me is not important coverage value itself but changes of coverage through time. Using the same engine every time is good enough for me.

milo:
Merged, thank you!

The reasons for preferred code coverage engines are:
1. PCOV - it is specialized extension for this purpose
2. PHPDBG - when someone runs Tester with PHPDBG, I assume that a main purpose is code coverage
3. Xdebug

PR #397, opened 2019-02-07
TestCaseRunner [WIP]

dg:
Can run multiple TestCase tests:

```php
/**
 * @testCase
 */

declare(strict_types=1);

namespace Tester;

require __DIR__ . '/../src/bootstrap.php';

(new TestCaseRunner)
	->findTests(__DIR__ . '/Test*.php')
	->run();
```

~~The next step could be removal of `@testcase`~~ … Done.

### DIFF
--- a/demo/TestFour.php
+++ b/demo/TestFour.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+
+class TestFour extends Tester\TestCase
+{
+	public function testMe()
+	{
+		Tester\Assert::true(true);
+		echo __FUNCTION__ . ',';
+	}
+
+
+	public function testMe2()
+	{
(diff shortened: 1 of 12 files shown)


dg:
Discussed here https://f3l1x.slack.com/messages/C69LKE8P2

JanTvrdik:
> Discussed here https://f3l1x.slack.com/messages/C69LKE8P2

That seems private.

f3l1x:
@JanTvrdik Wanna join us?

hrach:
> @JanTvrdik Wanna join us?

No, I want to see just the answer to Jan Tvrdik's question ;)

mabar:
It should make possible to separate test definition from it's execution so tests could be e.g. analysed with phpstan.

Issue #350, opened 2017-02-07
Assert::match does not match valid regex


PunchyRascal:
- bug report? yes/no
- feature request? yes/no
- version: v1.7.1

### Description
Until you realize, that your regex pattern cannot be delimited with `/`, the function does not work properly.

### Steps To Reproduce
fails: `Assert::match('/^foo$/', 'foo');`
passes: `Assert::match('#^foo$#', 'foo');`

Perhaps it would be better to avoid this magic behavior and add an extra assert for those tester expressions  (with appropriate name) and keep this only for regular expressions, to avoid confusion. Or vice versa.

milo:
Maybe add `Assert::matchRE()` and `Assert::outputMatchRE()`?

PunchyRascal:
Yeah, sure. 

Sorry for the delay, I am now browsing through all my open issues, must have missed it back then

PR #398, opened 2019-02-17
Better Assert::equal() WIP

dg:
- new feature
- BC break? no

This PR attempts to simplify the comparsion of structures that contain some variable data like this:

```php
$actual = ['a' => md5($s), 'b' => new DateTime, 'c' => rand()];
$expected = ???
Assert::equal($expected, $actual);
```
I'd like make assertions that 
- `$actual` is array
- it contains keys `a`, `b`, `c`
- and first value is hexa string: `Assert::type('string')` + `Assert::match('%h%', $value)`
- second value is object DateTime: `Assert::type('DateTime', $value)`
- the third value is integer: `Assert::type('int', $value)`) and >= 0 (`Assert::true($value >= 0)`

This PR allows to describe it using Assert::equal()  + language similar to how `Assert` is used. Only `Assert::` is replaced with `Value::`

```php
$actual = ['a' => md5($s), 'b' => new DateTime, 'c' => rand()];
$expected = [
	'a' => Value::type('string')->and(Value::match('%h%')),
	'b' => Value::type('DateTime'), 
	'c' => Value::type('int')->and(function ($val) { return $val >= 0; }),
];
Assert::equal($expected, $actual);
```

In case of error it writes:

```
Failed: ['a' => 'd41d8cd98f00b204e9800998ecf8427e', 'b' => DateTime(2019-02-17 22:36:36 +0100)(#532c), ...] should be equal to
    ... ['a' => type(string),match(%h%), 'b' => type(DateTime), 'c' => type(int),custom()]
```

Class name `Value` and methods are just working names. Is `Assert::equal()` good choice?

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -72,7 +72,7 @@ class Assert
 
 
 	/**
-	 * Asserts that two values are equal. The identity of objects,
+	 * Asserts that two values are equal and checks expectations. The identity of objects,
 	 * the order of keys in the arrays and marginally different floats are ignored.
 	 */
 	public static function equal($expected, $actual, string $description = null): void
@@ -85,13 +85,17 @@ class Assert
 
 
 	/**
-	 * Asserts that two values are not equal. The identity of objects,
+	 * Asserts that two values are not equal and checks expectations. The identity of objects,
 	 * the order of keys in the arrays and marginally different floats are ignored.
 	 */
 	public static function notEqual($expected, $actual, string $description = null): void
 	{
 		self::$counter++;
-		if (self::isEqual($expected, $actual)) {
+		try {
+			$res = self::isEqual($expected, $actual);
+		} catch (AssertException $e) {
+		}
+		if (empty($e) && $res) {
 			self::fail(self::describe('%1 should not be equal to %2', $description), $actual, $expected);
 		}
 	}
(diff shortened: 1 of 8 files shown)


src/Framework/Value.php
```diff
@@ -0,0 +1,72 @@
...
+
+
+	public function __call(string $name, array $args): self
+	{
+		$this->tests[] = [$name, $args];
+		return $this;
+	}
+
+
+	public function assert(callable $cb): self
```
JanTvrdik: must require `Closure` to avoid collision with `__call`

JanTvrdik:
Maybe call it `Constraint` or `Rule`? Mockery calls it [Matcher](https://github.com/mockery/mockery/tree/1.2.2/library/Mockery/Matcher).

JanTvrdik:
I think that it's quite problematic that both `__call` and `__callStatic` accept the same method name, making it probably impossible to document with `@method` annotation.

dg:
Or what about `Is` for readability?

```php
Is::notSame($expected)
Is::equal($expected)
Is::notEqual($expected)
Is::contains($needle) // hmmm
Is::notContains($needle)
Is::nan()
Is::truthy()
Is::falsey()
Is::count(int $count)
Is::type($type)
Is::match(string $pattern) // hmmm..
Is::matchFile(string $file)
```

These ones will probably will never used:
```php
Is::same($expected)
Is::true()
Is::false()
Is::null()
```

dg:
>  I think that it's quite problematic that both `__call` and `__callStatic` accept the same method name, making it probably impossible to document with `@method` annotation.

In fact, bigger problem is that `Is::abc()::xyz()` is perfectly valid, but means something different, `Is::xyz()`.

dg:
So better will be this syntax `Value::type('string')->and(Value::match('%h%')),`

milo:
👍 What about `Expect` class name?

milo:
@dg You mentioned `::pattern()` It is related to #350. I'm planning to split `Assert::match()` to `::matchRe()` or `::regexp()` or something like this.

milo:
And what about implement all static `Assert` on `Value` class instead of `__callStatic()`? Stack trace will be cleaner.

dg:
'Expect' sound good.

> And what about implement all supported static `Assert` methods on `Value` class instead of `__callStatic()`? Stack trace will be cleaner.

This is of course possible, but where do you see the stack trace?

milo:
> This is of course possible, but where do you see the stack trace?

An imaginary stack trace when exception thrown somehow in Assert. Small detail only.

PR #399, opened 2019-02-19
Assert::equal() makes the difference in order of lists (BC break)

dg:
- BC break? yes

It affects only seqences indexed from. Here is the anomaly in sorting usually undesirable. So with this change these arrays are not equal anymore:

```php
['a', 'b']
[1 => 'b', 0 => 'a'] 
```

It is BC break. I would suggest merge it in the master **after** releasing the next version to see if it hasn't a negative impact.


milo:
:+1:

PR #400, opened 2019-02-19
Code coverag engine name instead of availability flag

milo:
For #396

CliRunner may do specific Code Coverage engine setup.


src/Runner/PhpInterpreter.php
```diff
@@ -107,9 +107,16 @@ public function getVersion(): string
...
 
+	/** @deprecated  */
 	public function canMeasureCodeCoverage(): bool
 	{
-		return $this->info->canMeasureCodeCoverage;
+		return $this->info->codeCoverageEngine !== null;
+	}
+
+
+	public function getCodeCoverageEngine(): ?string
```
dg: What about getCodeCoverageEngines(): array ?

milo: And choose the one in CliTester? :+1:

dg: Yes.

PR #402, opened 2019-03-15
Prints the result of the current test

dg:
- new feature?  
- BC break? no

It may not be clear from the blank screen is test was successful. Did it start? Didn't PHP crashed (parse error?) before Environment::setup() was called? Did I started the test at all or I see a few minutes old result in the console? (Sometimes I am tired :)

### DIFF
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -39,6 +39,9 @@ class Environment
 	/** @var int initial output buffer level */
 	private static $obLevel;
 
+	/** @var int */
+	private static $exitCode = 0;
+
 
 	/**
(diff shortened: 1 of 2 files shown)


milo:
Thank you! Merged 80dfec31a3e05b91e511e982461cee68a9c291ed I only fixed (https://github.com/nette/tester/commit/80dfec31a3e05b91e511e982461cee68a9c291ed#diff-107fa28ddf64e59014ab89c08998ed7cR160) the Windows/AppVeyor situation with escape sequences for colors. It contains `[` char and test case methods listing didn't work.

PR #407, opened 2019-08-05
Performance assert

janbarasek:
- new feature
- BC break? no

New very useful assert `Assert::performance()` added for define maximum time limit for function execution time.

Params: `performance(float $limitMs, callable $function, string $description = null)`.

Example:

```php
Assert::performance(300, function () {
	(new Passwords)->hash('secret');
}, 'Generating must be faster than 300 ms.');
```

I think this basic assert should be included in native Tester to define boundaries before the start of development.


dg:
The duration of the test is not stable over time, please solve it in your code.

PR #410, opened 2019-08-20
readme: added with()

### LINKED ISSUE #409: Annotation like @accessPrivateMethod
Very often I need to test private methods in packages, models, ...

... but I can not.

Imagine test:

```php
class QueryNormalizerTest extends TestCase
{

	/**
	 * @var QueryNormalizer
	 */
	private $queryNormalizer;

	/**
	 * Expected syntax:
	 * @accessPrivateMethod $queryNormalizer::taskFixBrackets
	 */
	public function testFixBrackets(): void
	{
		Assert::same('1+2*(9-4)', $this->queryNormalizer->taskFixBrackets('1+2*(9-4'));
	}

}
```

QueryNormalizer class:

```php
class QueryNormalizer
{

	public function normalize(string $query): string
	{
		return ...;
	}

	private function taskFixBrackets(string $query): string
	{
		return ...;
	}

}
```

Currently I can not test private methods. I know I can call public method in class what will call some private methods, but in real case it is so hard. I want split tests and make separately logic for independent atomic test for specific methods and "big" test for call all logic like user of class.

When I can not test methods separately it's extremely hard to debug.

### Possible solution

Define new annotation `@accessPrivateMethod` to test where I can define name of property with class (service) instance and specific method name.

For example:

`@accessPrivateMethod $queryNormalizer::taskFixBrackets`

In TestCase will be inner logic for `ReflectionMethod` what change method to pubic by `setAccessible` or other magic.

If my suggestion make sense I will be very glad for implement it. :)

Thanks.

janbarasek:
- bug fix #409 
- BC break? no

Added documentation how call private method.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,6 +111,7 @@ This table shows all assertions (class `Assert` means `Tester\Assert`):
 - `Assert::match($pattern, $value)` - Compares result using regular expression or mask.
 - `Assert::matchFile($file, $value)` - Compares result using regular expression or mask sorted in file.
 - `Assert::count($count, $value)` - Reports an error if number of items in $value is not $count.
+- `Assert::with($objectOrClass, $closure)` - Executes function that can access private and protected members of given object via $this.
 
 Testing exceptions:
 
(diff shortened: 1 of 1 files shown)


dg:
Thanks

PR #411, opened 2019-09-09
Environment: File can be undefined.

janbarasek:
- bug fix
- BC break? yes

In case of Doctrine ORM Schema update when all entities are loaded automatically, `$trace` with index `file` can be undefined.

Snímek obrazovky 2019-09-09 v 13 54 22

### DIFF
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -166,8 +166,8 @@ class Environment
 	public static function getTestAnnotations(): array
 	{
 		$trace = debug_backtrace();
-		$file = $trace[count($trace) - 1]['file'];
-		return Helpers::parseDocComment(file_get_contents($file)) + ['file' => $file];
+		$file = $trace[count($trace) - 1]['file'] ?? null;
+		return Helpers::parseDocComment($file ? file_get_contents($file) : '') + ['file' => $file];
 	}
 
 


dg:
Thanks

dg:
What is the reason that there is not `file`? Could you dump the trace here?

janbarasek:
@dg In case of:

```php
/**
 * Returns current test annotations.
 */
public static function getTestAnnotations(): array
{
	$trace = debug_backtrace();
	$file = $trace[count($trace) - 1]['file'] ?? null;
	if ($file === null) {
		echo \Tracy\Dumper::toTerminal($trace);
	}
	return Helpers::parseDocComment($file ? file_get_contents($file) : '') + ['file' => $file];
}
```

First part looks ok:

Snímek obrazovky 2019-09-09 v 17 54 29

But last part not:

Snímek obrazovky 2019-09-09 v 17 55 23

Source:

Snímek obrazovky 2019-09-09 v 17 58 15

Source code: https://github.com/baraja-core/doctrine/blob/master/src/OrmSchemaUpdateTool.php#L24

Called command in Terminal: `php www/index.php o:s:u -f --dump-sql`

I think the bug is in my TestCase class, because I should use `if (isset(...` in this construction here:

```php
if (isset($_SERVER['NETTE_TESTER_RUNNER'])) {
	$di = Booting::bootForTests()->createContainer();

	(new QueryToLatexTest(
		$di->getByType(Tokenizer::class),
		$di->getByType(QueryNormalizer::class)
	))->run();
}
```

I think it's relatively complex problem. Do you understand the principle?

dg:
And what is in `www/index.php`? How is `OrmSchemaUpdateTool::run()` called?

janbarasek:
@dg I use Nette sandbox:

```php
<?php

declare(strict_types=1);

require __DIR__ . '/../vendor/autoload.php';

App\Booting::boot()
	->createContainer()
	->getByType(Nette\Application\Application::class)
	->run();
```

But in `Container` in `initialize()` method is (in Container specific for CLI):

```php
OrmSchemaUpdateTool::setContainer($this);
register_shutdown_function([OrmSchemaUpdateTool::class, 'run']);
```

When I think about it, it wasn't very smart. I use it as a hack to call a specific piece of code in CLI mode.

dg:
So it is called via `register_shutdown_function()`, thats why `file` is missing.

janbarasek:
I think yes.

PR #412, opened 2019-09-22
Dumper: keep full data provider name in output files

vojtech-dobes:
- new feature
- BC break? no 

I created test case `Test.phpt` with following data provider variations:
- `source_1`
- `source_2`
- `source_3`

All of them failed on assert, but only single output pair of files was produced:
```
Test.[dataprovider=source-].actual
Test.[dataprovider=source-]
```

The number got cut off due to limitations of 20 chars in `Dumper` - which this PR changes to unlimited amount until encountering `|` which seems to be used as delimiter by Tester.

Please let me know if this is reasonable improvement; from my perspective it would help in my case tremendously (`source_<n>` is supposed to be actually meaningful longer thing).


milo:
Merged 76cef6c61b4ca80eb4dbdacb5d9efc3e4b6feb2c

PR #416, opened 2019-11-22
Allow SVG elements in DOM (close #354).

### LINKED ISSUE #354: DomQuery does not support SVG
Another DOMDocument-caused bug. DomQuery does not support SVG.
Current workaround I use is `Strings::replace($html, '#<svg[^>]*+>.+?</svg>#s', '');`

Our SVG code

~~~html
<svg role="img" aria-label="" title="" class="shape shape-share-fb view-head button-icon">
	<use xlink:href="#shape-share-fb"></use>
</svg>
~~~

smuuf:
- bug fix
- BC break? no

Added `svg` to the list of supported HTML tags. Also added the rest of known SVG elements as per [MDN SVG Elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element) page.

### DIFF
--- a/src/Framework/DomQuery.php
+++ b/src/Framework/DomQuery.php
@@ -37,7 +37,14 @@ class DomQuery extends \SimpleXMLElement
 		libxml_use_internal_errors($old);
 
 		$re = '#Tag (article|aside|audio|bdi|canvas|data|datalist|figcaption|figure|footer|header|keygen|main|mark'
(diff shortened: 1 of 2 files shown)


dg:
There is also a lot of SVG elements https://developer.mozilla.org/en-US/docs/Web/SVG/Element

genesiscz:
@smuuf are you gonna complete this one or not?

smuuf:
I added the rest of known SVG elements as per the MDN SVG elements page + some tests.

dg:
Thanks. When I think about it, it may be better to allow all the elements. Because there are MathML and Custom Elements…

Issue #401, opened 2019-02-21
Class 'SimpleXMLElement' not found


milo:
DomQuery extends \SimpleXMLElement which is missing when dom extension not loaded. Fatal error is thrown when loading `bootstap.php` (no autoload is used).

PR #419, opened 2020-06-27
simplexml must be required #401

villfa:
- bug fix / new feature? Fix issue #401
- BC break? no

The `\SimpleXMLElement` class is used under the `src/` directory (see below) so the `simplexml` extension must be required to avoid a *class not found* error.

https://github.com/nette/tester/blob/0b33a936d28faa641f816b842020ae50f85ec6c3/src/Framework/DomQuery.php#L16

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,10 @@
 		}
 	],
 	"require": {
-		"php": ">=7.1"
+		"php": ">=7.1",
+		"ext-simplexml": "*"
 	},
 	"require-dev": {
-		"ext-simplexml": "*",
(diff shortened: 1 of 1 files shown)


villfa:
@dg I understand that my proposition doesn't satisfy you (I guess you don't want `ext-simplexml` to be required for installing the library) but the 170b379 commit won't fix the issue.

The condition cannot be evaluated since the `class not found` error happens when the class is declared.

Another solution could be to move this condition to `bootstrap.php`.

dg:
Oh yes! I used your solution with bootstrap.php

PR #421, opened 2020-08-16
add global function test()

dg:
This is similar to the `test()` function, which is actually used by all Nette libraries. In addition, the test name is passed to the function and it lists results.

image
(https://github.com/nette/utils/blob/be6f0157dbb500c90ad0762a32a0681e087a6be3/tests/Utils/Arrays.insertBefore().phpt#L24-L72)

It also adds the `beforeEach()` and `afterEach()` functions.

What to consider
- The function is only created if it does not exist, which should prevent collisions. Maybe it could be controlled in any other way.
- switch whether to continue to the next `test()` function if assertion fails
- switch to turn off output
- output directly to STDOUT, so it cannot then be controlled by an output buffer
- consider using the TestCase for before/after execution, similary to setUp/tearDown

### DIFF
--- a/src/Framework/Environment.php
+++ b/src/Framework/Environment.php
@@ -36,6 +36,9 @@ class Environment
 	/** @var bool */
 	public static $useColors;
 
+	/** @var ?TestFunction */
+	public static $testFunction;
+
 	/** @var int initial output buffer level */
 	private static $obLevel;
 
@@ -50,6 +53,7 @@ class Environment
 	{
 		self::setupErrors();
 		self::setupColors();
+		self::setupFunctions();
 		self::$obLevel = ob_get_level();
 
 		class_exists(Runner\Job::class);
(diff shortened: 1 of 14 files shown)


milo:
Uff, sorry, I just wated to send comment, not to start review. Just clicked the first green button I saw :)

milo:
I'm in 👍 

I proposed similar thing, `Assert::block(callable ...)`, in the past, but I cannot find it now. And I can't recall why we didn't use it :o)

milo:
> switch to turn off output

I think this can be controlled by the same option as color output is.

milo:
> The function is only created if it does not exist, which should prevent collisions. Maybe it could be controlled in any other way.

Just thinking a loud... What about to do not load it in `Environment::setup()` but let user call explicitly `Envitonemt::setupFunctions()`? When coparing to Tracy's `dump()`, it just works fine without conflict, but `dump()` calls do not remain in code. On the other hand, tests are not prouduction evironment as wall, so probably detection is OK.

milo:
> switch whether to continue to the next test() function if assertion fails

Something like deferred fail?

Maybe always when run directly by `php test.phpt` and not by Runner?

PR #422, opened 2020-09-07
Upgraded .html report for code coverage

mesour:
- improved existing feature
- BC break? /no
- doc PR: needs only new example of coverage report to docs - [nette/utils coverage.zip](https://github.com/nette/tester/files/5184997/coverage.zip)

Loading in the browser is about one order of magnitude faster (from 3 sec it is now 0.3 sec) and the size of the generated file is about 20% smaller. (speed and size tested on a report of 1000 files with 110k lines).

Added feature to sort results by% coverage and location has been added, a new tab with display the tree of files in folders and% coverage for every folder.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,5 +2,9 @@
 .gitignore export-ignore
 .github export-ignore
 .travis.yml export-ignore
+ecs.php export-ignore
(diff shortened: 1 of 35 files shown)


dg:
That's a great improvement!

PR #424, opened 2020-12-03
CodeCoverage: better support for Xdebug 3

jiripudil:
- bug fix
- BC break? no
- doc PR: not necessary

[Xdebug 3](https://xdebug.org/docs/upgrade_guide) requires `xdebug.mode` to include `coverage` for coverage to work, otherwise coverage-related symbols are not defined and tests fail with `Error: Undefined constant "Tester\CodeCoverage\XDEBUG_CC_UNUSED"`.

Users can make it work by running tester with `-d xdebug.mode=coverage`, but I thought tester could setup this automatically, so that it works out of the box as with Xdebug 2. Therefore, this PR adds version number to code coverage engine info and uses that to automatically setup `coverage` mode for test processes when Xdebug >= 3.0.0 is used.

### DIFF
--- a/src/CodeCoverage/Collector.php
+++ b/src/CodeCoverage/Collector.php
@@ -30,9 +30,9 @@ class Collector
 	public static function detectEngines(): array
 	{
 		return array_filter([
-			extension_loaded('pcov') ? self::ENGINE_PCOV : null,
-			defined('PHPDBG_VERSION') ? self::ENGINE_PHPDBG : null,
-			extension_loaded('xdebug') ? self::ENGINE_XDEBUG : null,
+			extension_loaded('pcov') ? [self::ENGINE_PCOV, phpversion('pcov')] : null,
+			defined('PHPDBG_VERSION') ? [self::ENGINE_PHPDBG, PHPDBG_VERSION] : null,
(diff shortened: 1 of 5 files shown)


src/CodeCoverage/Collector.php
```diff
@@ -30,9 +30,9 @@ class Collector
 	public static function detectEngines(): array
 	{
 		return array_filter([
-			extension_loaded('pcov') ? self::ENGINE_PCOV : null,
-			defined('PHPDBG_VERSION') ? self::ENGINE_PHPDBG : null,
-			extension_loaded('xdebug') ? self::ENGINE_XDEBUG : null,
+			extension_loaded('pcov') ? [self::ENGINE_PCOV, phpversion('pcov')] : null,
+			defined('PHPDBG_VERSION') ? [self::ENGINE_PHPDBG, PHPDBG_VERSION] : null,
+			extension_loaded('xdebug') ? [self::ENGINE_XDEBUG, phpversion('xdebug')] : null,
```
JanTvrdik, line 35: maybe `[self::ENGINE_XDEBUG => phpversion('xdebug')]`?

dg: It would be harder to work with.

JanTvrdik: i thought it would be easier xd

PR #425, opened 2020-12-03
fix Assert::type() failure error when expected $type is object

jiripudil:
- bug fix
- BC break? no
- doc PR: not necessary

Apparently `Assert::type()` supports an object instance in place of expected `$type`, but if such assertion fails, building a failure message produces either an `Error: Object of class (...) could not be converted to string`, or a rather misleading message if the expected type implements `__toString`.


dg:
Thanks

PR #426, opened 2020-12-08
Measure test run time duration

milo:
- new feature
- BC break? no

It is a quite hard to measure test run time duration precisly. It is possible from test itself, but things get complicated with `@testCase` and `@provider` annotations when collecting/pairing durations back to custom output handler.

I think that this kind of measurement is quite precise and easy to access for OutputHandler.

Time duration is unnecessary most of the time, but it is very useful when reducing total amount of time with higher quantity of tests (100+). For examle, database locks or php.ini tunning (e.g. loading SNMP extension eats hundreds of miliseconds).

### DIFF
--- a/src/Runner/Job.php
+++ b/src/Runner/Job.php
@@ -55,6 +55,9 @@ class Job
 	/** @var string[]  output headers */
 	private $headers = [];
 
+	/** @var float|null */
+	private $duration;
+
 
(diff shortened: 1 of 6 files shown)


milo:
Was thinking about `withDuration()` method but duration should be known when setting test result.

PR #427, opened 2020-12-17
Add hasKey and hasNotKey assertions

KminekMatej:
- new feature
- BC break? no
- doc PR: nette/docs#???  On its way

Add simple assertions hasKey nad hasNotKey to tester. They are very useful for testing restful api responses.

Tests included.


dg:
Thanks

PR #428, opened 2021-01-03
Fix integer-indexed cases

vojtech-dobes:
- **bug fix**
- BC break? *not sure to be honest*
- doc PR: *shouldn't be needed*

### Description
Currently when PHP @dataProvider returns integer-indexed cases (simply array), in each run the first case is used.

### Example
Given following test:
```php
[$expected, $actual] = Tester\Environment::loadData();
Tester\Assert::same($expected, $actual);
```

With following data provider, all 3 cases will fail:
```php
return [
    ['foo', 'bar'],
    ['foo', 'foo'],
    ['foo', 'foo'],
];
```

With following data provider, all 3 cases will succeed:
```php
return [
    ['foo', 'foo'],
    ['foo', 'bar'],
    ['foo', 'bar'],
];
```

### Summary
That's what this PR fixes :)

### DIFF
--- a/src/Framework/DataProvider.php
+++ b/src/Framework/DataProvider.php
@@ -44,7 +44,7 @@ class DataProvider
 		}
 
 		foreach ($data as $key => $value) {
-			if (!is_int($key) && !self::testQuery($key, $query)) {
+			if (!self::testQuery((string) $key, $query)) {
 				unset($data[$key]);
 			}
 		}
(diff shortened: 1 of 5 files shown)


milo:
Uff, this is nasty bug. Thank you!

milo:
The `is_int()` condition is there because I thought, integer indexed items should never be filtered out. The bug is hidden in here https://github.com/nette/tester/blob/43743fb37aa1141bcacefb3481b8dd0bc1e16be2/src/Framework/Environment.php#L230

I have to thing about it.

vojtech-dobes:
@milo Awesome, thanks!

PR #429, opened 2021-01-06
CliTester: added missing newline

diegosardina:
- bug fix
- BC break? no

Added new line at the end to avoid this:
Generating code coverage report... failed. Coverage file is empty. Do you call Tester\Environment::setup() in tests?diego@localhost $


dg:
Thanks

PR #431, opened 2021-04-06
TestCase::prepareTestData(): Data provider validation must be inside foreach.

janbarasek:
- bug fix
- BC break? no

Validation must be inside foreach, because this code is not valid input, but original will ignore it:

Snímek obrazovky 2021-04-06 v 21 39 44

Thanks.


janbarasek:
I am not sure how to all tests for this bugfix. :/

PR #432, opened 2021-04-09
Dumper: create output directory recursively

forrest79:
- new feature
- BC break? no

Just create dumper output directory recursively - it's handy when someone use a custom directory path for output files.


milo:
I'm not sure about this issue. I understand the need, but recursive `mkdir` is not thread safe.

forrest79:
Thanks for the response, I didn't know about this (I already find PHP bug request for this). So let's close this, we can solve this manually.

PR #438, opened 2022-07-10
Capture stderr to temp file and output it in test results

MartinMystikJonas:
- new feature
- BC break: no (only additional output in test results)

### DIFF
--- a/src/Runner/CliTester.php
+++ b/src/Runner/CliTester.php
@@ -57,6 +57,9 @@ class CliTester
 
 		if ($this->options['--info']) {
 			$job = new Job(new Test(__DIR__ . '/info.php'), $this->interpreter);
+			if ($this->options['--temp'] !== null) {
+				$job->setTempDirectory($this->options['--temp']);
+			}
 			$job->run();
 			echo $job->getTest()->stdout;
 			return null;
--- a/src/Runner/Job.php
+++ b/src/Runner/Job.php
@@ -27,9 +27,10 @@ class Job
 	/** waiting time between process activity check in microseconds */
 	public const RUN_USLEEP = 10000;
 
-	public const
-		RUN_ASYNC = 1,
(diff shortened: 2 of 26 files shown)


MartinMystikJonas:
Any idea why output tests on my computer works just fine but fails in GitHub actions? Output does not match but it seems it should. I suspect some whitespace difference but cannot find it.

MartinMystikJonas:
I still struggle to find what causes tests to fail only in CI. I identified that it is caused by spaces only lines in actual output on CI that does not happens on by local build. But I have no idea why. I will do some more experimentation to find it.

MartinMystikJonas:
Finall I made it pass. I also fixed one coding standard error.

@dg @milo Could oyu check it out and let me know what do you think?

milo:
Looks great. I'll polish it a little bit and I'll merge it.

milo:
@MartinMystikJonas Are you OK with such modifications? https://github.com/milo/tester/tree/MartinMystikJonas-stderr-file

MartinMystikJonas:
@milo Looks OK

MartinMystikJonas:
@milo Is there anything I should do now to finish this PR?

milo:
Thank you. I've merged it.

dg:
Do you have any idea why the `RunnerOutput\OutputHandlers.phpt` test is not passing on Windows? Difference is in progress line:

actual:

```
PHP 8.2.0beta3 (cgi-fcgi) | …

FFFF...ssss
```

expected:
```
PHP 8.2.0beta3 (cgi-fcgi) | …

F.sF.sFsF.s
```

MartinMystikJonas:
I ajve seen something similar when temp dir was not purged between tests so order of tests prioritized failed ones. Why it happens on windows I do not know.

dg:
Bingo, thanks!

PR #443, opened 2023-08-12
Added `console-lines` output mode which prints each test on a separate line

smuuf:
Usage example:
```sh
$ ./src/tester -C -o console-lines tests/
```

This is handy for environments with "non-standard" _(buffered, I guess?)_ handling of standard output, for example `Github Actions` _(where a progress of tests cannot be seen until until end-of-line appears, which in standard `console` mode happens only when all tests finish)_ or `Docker Compose` logging output _(where in standard `console` mode each finished test's dot is printed alone on separate line)_.

Or the `console-lines` mode can be handy just to see a more detailed progress of tests in all environments, because it outputs something like this:

<table>
<tr>
<td>Passing tests</td>
<td>Failing tests</td>
<td>Usage</td>
<td>Also supports "cider mode"</td>
<tr/>
<tr>
<td>

obrazek

</td>
<td>

obrazek

</td>
<td>

obrazek

</td>
<td>

obrazek

</td>
<tr/>
</table>

Also, "cider mode" now shows a lemon emoji for skipped tests.

- ***new feature***
- BC break? ***no***
- doc PR: https://github.com/nette/docs/pull/1027

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,6 @@
 parameters:
 	level: 5
-
 	paths:
 		- src
+	typeAliases:
+		Alias_TestResultState: 'Tester\Runner\Test::Passed|Tester\Runner\Test::Skipped|Tester\Runner\Test::Failed|Tester\Runner\Test::Prepared'
--- a/readme.md
+++ b/readme.md
@@ -223,7 +223,8 @@ Options:
     -s                           Show information about skipped tests.
     --stop-on-fail               Stop execution upon the first failure.
     -j <num>                     Run <num> jobs in parallel (default: 8).
-    -o <console|tap|junit|none>  Specify output format.
+    -o <console|console-lines|tap|junit|none>
+                                 Specify output format.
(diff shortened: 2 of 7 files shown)


dg:
I like that! I'd probably just use a slightly different name, like `-o verbose`?

smuuf:
I'd think `verbose` could mean different things for different output modes, so maybe usurping the word `verbose` for something related just to the `console` output might not be the best thing to do, future-wise, but other than that I have no strong opinion on that 😄

If you'd prefer `verbose`, I can make it be called `verbose`.

smuuf:
Updated: Renamed to `verbose` + updated the relevant [PR to docs](https://github.com/nette/docs/pull/1027).

dg:
I was just throwing that out there, I'm sure someone can think of something better.

MartinMystikJonas:
Imho 'console-lines' was better. Or just 'lines'?

Issue #448, opened 2024-05-17
Display name of concrete test method in console-lines output


Marian-Kechlibar:
Nette Tester currently supports a new output format console-lines, which looks very nice and, IMHO, is the most informative of them all. An optimal combination of colors and information, as far as my human eyes go.

What I nevertheless miss is the names of individual methods that are running. The "log" output format has them, but is completely grey and thus less informative for a human to look at. The "console-lines" format doesn't have them.

As a result, the output is less informative:

· 1/12 Admin\CertificateTest.php FAIL in 0.00 s
· 2/12 Admin\CertificateTest.php FAIL in 0.00 s
· 3/12 User\CertificateTest.php FAIL in 0.00 s
· 4/12 Admin\CertificateTest.php OK in 1.33 s
· 5/12 Admin\CertificateTest.php OK in 0.71 s
· 6/12 Admin\CertificateTest.php OK in 0.70 s
· 7/12 Admin\CertificateTest.php OK in 0.34 s
· 8/12 Admin\CertificateTest.php OK in 0.34 s
· 9/12 Admin\CertificateTest.php OK in 0.39 s

I propose a wider format like this

· 1/12 Admin\CertificateTest.php (testCertificateList) FAIL in 0.00 s

smuuf:
This is something I've also been wanting to add, so I finally did it. Thanks for the nudge 🐣

PR #451, opened 2024-06-12
Add testcase method's name to the test's title

smuuf:
Resolves https://github.com/nette/tester/issues/448.

- new feature
- BC break? no (hopefully)

## What this does

For tests wrapped in testcase class:
```php
class MyTests extends \Tester\Testcase {
    public function testMethod_1() { ... }
    public function testMethod_2() { ... }
}
```

... where Tester creates a separate `Test` object for each of the testcase's methods, ***we now add (append) the method's name to the (maybe already specified) test's title***.

In the case of these two methods, if this PR is merged, the first test object will now have the title `"testMethod_1"` and the second one will have the title `"testMethod_2"`.

## What is looks like
... if Nette Tester self-tests with [`-o console-lines` output mode](https://github.com/nette/tester/pull/443):
obrazek

## Append... title?
Yes, the test file might also contain "a primary, standalone title definition", like so:

```php
/**
 * These are my tests
 */
class MyTests extends \Tester\Testcase {
    public function testMethod_1() { ... }
    public function testMethod_2() { ... }
}
```

... in that case the test objects representing both methods would have a same, pre-specified title `"These are my tests"`.

This PR changes this and in the end the first test object would have the title `"These are my tests testMethod_1"` and the second one will have the title `"These are my tests testMethod_2"`.

### DIFF
--- a/src/Runner/Test.php
+++ b/src/Runner/Test.php
@@ -43,7 +43,7 @@ class Test
 	public function __construct(string $file, ?string $title = null)
 	{
 		$this->file = $file;
-		$this->title = $title;
+		$this->title = $title !== null ? trim($title) : null;
 	}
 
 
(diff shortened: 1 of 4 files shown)


src/Runner/Test.php
```diff
@@ -43,7 +43,7 @@ class Test
 	public function __construct(string $file, ?string $title = null)
 	{
 		$this->file = $file;
-		$this->title = $title;
+		$this->title = $title !== null ? trim($title) : null;
```
dg, line 46: In what case is the constructor called with a title that has extra spaces?

smuuf:
Thanks for merging 🌻  Unfortunately your changes introduced a minor bug with dealing with spaces in the test's title: 

https://github.com/nette/tester/commit/7d226fa4f53515e62642c97e0dddcdf8bd7e4a19#diff-07c33aaa3cca21e4927002dc11edb6d232627948635ad1e537854703ecd21666R231-R233

```php
	return array_map(
			fn(string $method): Test => $test
				->withTitle("$test->title $method")
				->withArguments(['method' => $method]),
			$methods,
		);
```

... If the initial instance of Test doesn't have any "primary" title (i.e. `$test->title` is `null` or `''`), then the new title will contain leading whitespace.

obrazek

It would probably be sufficient if it was done as `->withTitle(trim("$test->title $method"))` instead.

PR #457, opened 2025-04-02
Dumper: print enum case directly in toPhp()

vojtech-dobes:
- new feature
- BC break? yes, I suppose

Reasoning: I was checking some test outputs and saw how enums get dumped, and was thinking if they were just simply inlined, the dump would be more readable.


dg:
Thanks

PR #462, opened 2025-11-25
Fix PHP 8.5 deprecation: null as array offset in -o option handler

JanTvrdik:
## Summary
- Fix PHP 8.5 deprecation warning when using `-o console` without a filename
- PHP 8.5 deprecated using `null` as an array offset; when no filename is specified, `$file` was `null` and used as key in `$outputFiles[$file]`
- Changed default from `null` to empty string `''` and updated corresponding checks

## Test plan
- [x] Verified all 88 tests pass with `src/tester tests -s -C`
- [x] Confirmed output handlers work correctly (they use `$file ?: 'php://output'` which handles empty string the same as null)


JanTvrdik:
Unlucky, `https://httpbin.org/` is down

PR #466, opened 2026-01-23
Add Countable to the count() doctype

spaze:
- bug fix / new feature?   bug fix, more or less
- BC break? no

Add `Countable` to the doctype, the native param type contains it as well.


dg:
thanks

PR #467, opened 2026-02-20
Feat/structural metrics v2 + DROP PHP8.0 support

### LINKED ISSUE #463: Work with lines outside class/struct in Clover XML report
Version: 2.5.7

### Bug Description
CodeCoverage generator, when specifying Clover XML as output natively does not work with lines outside class / struct definition at project metrics level.

### Steps To Reproduce
Run local tests of this project. Tested file `tests/CodeCoverage/fixtures.clover/Logger.php` has two lines outside the class scope:

```
$logger = new Logger('php://stdout');  # 1
$logger->log('foo');  # 1
```
These lines has been tested and reported by xdebug as covered, but `statementsCount` metric did not count them:
```
statements: 5
coveredStatements: 3
```

### Expected Behavior
```
statements: 7
coveredStatements: 5
```

### Possible Solution
PR In the comments

KminekMatej:
This PR created as a replacement of https://github.com/nette/tester/pull/464 - to avoid complicated rebase process

bug fix https://github.com/nette/tester/issues/463 https://github.com/nette/tester/issues/463
BC break? no
doc PR: Not neccessary
In my project I had to manually count code coverage for further reporting. To avoid parsing the code coverage from standard HTML report I wanted to use machine-ready structure so I checked Clover XML output. I expected by dividing project->metric->coveredStatements / project->metric->statements to reach the same percents as HTML output has, but surprisingly the number was quite different.

During diving into the used algorithm I found out that Clover XML report does not take into account lines which are actually present in the file outside the class / trait scope. Even with a small todo mentioned in comments that it needs to be finished. So I tried to finish it.

During analyzation of the code, there is new function which counts the stats of the lines outside classes / traits and adds these stats into project metrics.

I believe this is smoething to be primarily checked by @milo , who (according to git blame) wrote most of the XML coverage code (big kudos 💪 ).

This PR does not fix the entire percent count completely, as the goal is to count fomr this report same percentil as is reported in the CLI Tester after the tests, but it gets it one step closer to the end.


KminekMatej:
Closing this — the functionality has since landed upstream independently.

  `CloverXMLGenerator` file and project metrics now include code outside classes as of
  69b2b69 ("_CloverXMLGenerator: file and project metrics include functions and code
  outside classes_"), with 4251f58 ("_PhpParser: recognizes enums and functions returning
  by reference_") covering enums and by-ref functions in the same area. The upstream
  approach counts free functions among methods and derives statements from all
  executable lines of a loaded file, which supersedes the `calculateStructuralMetrics()`
  approach proposed here.

  The remaining commits on this branch are also obsolete against current master:

  - Dropping PHP 8.0 — master already declares `"php": "8.1 - 8.6"` (0ac919f)
  - Coding style (`0o` octal literals) — already applied via 62103c5 (dresscode)

Hope at least my PR served as an suggestion about the fixed needed to implement :).

  Thanks for the good job @dg !

PR #468, opened 2026-03-27
PhpInterpreter: pass command to proc_open() as array

milo:
I noticed in PHP manual that `$command` to `proc_open()` can be passed as an array. In that case:
- PHP escapes arguments itself (seems more secure)
- subshell call is bypassed on Linux too (micro speed-up)


dg:
Thanks

